PR #32, opened 2016-02-11
`vagrant service-manager env` return all info

bexelbie:
Today that means just Docker, but it should work :)


navidshaikh:
LGTM!

PR #35, opened 2016-02-11
Add notice when copying certificates

### LINKED ISSUE #4: Don't use .docker for Docker client certificates
_From @hferentschik on January 28, 2016 12:9_

Currently the Docker client certificates end up in a hidden directory `.docker`, eg `.vagrant/machines/cdk/virtualbox/.docker/`. It seems odd to have another hidden directory underneath `.vagrant`. Just `docker` would be better.

_Copied from original issue: projectatomic/vagrant-adbinfo#80_

### LINKED ISSUE #5: Info log messages should be used when certificates get copied to host
_From @hferentschik on January 28, 2016 12:3_

Currently everything adbinfo does is in silent mode which makes it hard to a user to understand what's even going on. For example it would be nice to get some info logging that client certificate files are getting copied to the host and to where. 

Atm the output of `vagrant adbinfo` needs to also be evaluated by a shell, so this might be hard to do. This is one more reason to make certificate generation and re-configuration of the Docker daemon part of the provisioning. 

Also the README as well as the built-in help should be more descriptive.

_Copied from original issue: projectatomic/vagrant-adbinfo#79_

bexelbie:
Fixes #5
Fixes #4

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -54,7 +54,7 @@ $vagrant service-manager env docker
         # this execute the operations needed to print the docker env info
         with_target_vms(nil, {:single_target=>true}) do |machine|
           # Path to the private_key and where we will store the TLS Certificates
-          secrets_path = machine.data_dir
+          secrets_path = File.expand_path("docker", machine.data_dir)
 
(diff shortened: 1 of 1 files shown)


navidshaikh:
LGTM! Excellent!

PR #37, opened 2016-02-11
Update README with quick start steps

navidshaikh:
Adds the reference to Vagrantfile since private networking is needed.


bexelbie:
LGTM

PR #42, opened 2016-02-12
Fixes the check for finding vagrant box state

### LINKED ISSUE #41: Plugin reports incorrect machine state
Even if vagrant box is running, plugin prompts to bring up the machine

```
➜  vagrant-service-manager git:(master) ✗ vagrant status
Current machine states:

default                   running (virtualbox)

The VM is running. To stop this VM, you can run `vagrant halt` to
shut it down forcefully, or you can run `vagrant suspend` to simply
suspend the virtual machine. In either case, to restart it again,
simply run `vagrant up`.
➜  vagrant-service-manager git:(master) ✗ bundle exec vagrant service-manager env docker
Your Gemfile lists the gem vagrant-service-manager (>= 0) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of just one of them later.
Vagrant appears to be running in a Bundler environment. Your 
existing Gemfile will be used. Vagrant will not auto-load any plugins
installed with `vagrant plugin`. Vagrant will autoload any plugins in
the 'plugins' group in your Gemfile. You can force Vagrant to take over
with VAGRANT_FORCE_BUNDLER.

You appear to be running Vagrant outside of the official installers.
Note that the installers are what ensure that Vagrant has all required
dependencies, and Vagrant assumes that these dependencies exist. By
running outside of the installer environment, Vagrant may not function
properly. To remove this warning, install Vagrant using one of the
official packages from vagrantup.com.

The virtual machine must be running before you execute this command.
Try this in the directory with your Vagrantfile:
    vagrant up

```

navidshaikh:
Fixes #41  This patch fixes the logic of finding if vagrant box is running

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -14,7 +14,7 @@ module Vagrant
       def execute
         # Check for running machines
         with_target_vms(nil, {:single_target=>true}) do |machine|
-            if machine.state.id != "running" then
+            if machine.state.id != :running then
                 message = <<-eos
 The virtual machine must be running before you execute this command.
 Try this in the directory with your Vagrantfile:


LalatenduMohanty:
LGTM , I have tested this and works for me.

brgnepal:
Fix working in my machine.

bexelbie:
If the Maintainers list has changed we need to update the file.

PR #43, opened 2016-02-12
Adds subcommand for displaying openshift information

### LINKED ISSUE #23: service-manager should also give OpenShift information
_From @LalatenduMohanty on November 17, 2015 11:36_

If the OpenShift is running in the ADB vagrant box, vagrant-adbinfo should publish the information

Refer: https://github.com/projectatomic/adb-atomic-developer-bundle/issues/110

_Copied from original issue: projectatomic/vagrant-adbinfo#26_

navidshaikh:
Fixes #23 
 This patch adds a subcommand in plugin to display the openshift information.
 Also if given `vagrant service-manager env` it prints the status of all the providers inside the box,
 at the moment it prints the status about docker and openshift.
 If openshift service is running inside the box, it prints the required connection information and exits with 0 status.
 If openshift service is not running [not present, disabled, stopped, activating, etc], it gives the error message and exits with 1 status.
 This implementation checks for openshift service via systemd service files.

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -32,10 +32,14 @@ module Vagrant
         when "env"
             self.exit_if_machine_not_running
             case subcommand
+
             when "docker"
                 self.execute_docker_info
+            when "openshift"
+                self.execute_openshift_info
             when nil
-                self.execute_docker_info
+                # display information about all the providers inside ADB/CDK
+                self.print_all_provider_info
             else
                 self.print_help
             end
@@ -76,6 +80,79 @@ $vagrant service-manager env docker
         end
       end
 
+      def check_if_a_service_is_running(machine, service)
+        with_target_vms(nil, {:single_target=>true}) do |machine|
+          machine.communicate.execute(command) do |type, data|
+          end
+        end
+      end
+
+
+      def print_all_provider_info
+        message = <<-msg
+# Showing the status of providers in the vagrant box:
+        msg
+        @env.ui.info(message)
+        self.execute_docker_info
+        self.execute_openshift_info
+      end
+
+      def execute_openshift_info
+        @@OPENSHIFT_PORT = 8443
+        command = "systemctl status openshift"
+        with_target_vms(nil, {:single_target=>true}) do |machine|
+          if machine.communicate.test(command) then
+            # Find the guest IP
+            guest_ip = self.find_machine_ip(machine)
+            openshift_url = "https://#{guest_ip}:#@@OPENSHIFT_PORT"
+            openshift_console_url = "#{openshift_url}/console"
+            self.print_openshift_info(openshift_url, openshift_console_url)
+          else
+            @env.ui.info("# OpenShift service is not running in the vagrant box.")
+            exit 1
+          end
+        end
+      end
+
+      def print_openshift_info(openshift_url, openshift_console_url)
+        if !OS.windows? then
+          message =
+          <<-eos
+# Set the following environment variables to access the OpenShift
+export OPENSHIFT_URL=#{openshift_url}
+export OPENSHIFT_WEB_CONSOLE=#{openshift_console_url}
+# run following command to configure your shell:
+# eval "$(vagrant service-manager env openshift)"
+          eos
+          @env.ui.info(message)
+        else
+          message =
+          <<-eos
+# Set the following environment variables to access the OpenShift
+setx OPENSHIFT_URL=#{openshift_url}
+setx OPENSHIFT_WEB_CONSOLE=#{openshift_console_url}
+          eos
+          @env.ui.info(message)
+        end
+      end
+
+      def find_machine_ip(machine)
+        # Find the guest IP
+        if machine.provider_name == :virtualbox then
+            # VirtualBox automatically provisions an eth0 interface that is a NAT interface
+            # We need a routeable IP address, which will therefore be found on eth1
+            command = "ip addr show eth1 | awk 'NR==3 {print $2}' | cut -f1 -d\/"
+        else
+            # For all other provisions, find the default route
+            command = "ip route get 8.8.8.8 | awk 'NR==1 {print $NF}'"
+        end
+        guest_ip = ""
+        machine.communicate.execute(command) do |type, data|
+          guest_ip << data.chomp if type == :stdout
+        end
+        return guest_ip
+      end
+
       def execute_docker_info
         # this execute the operations needed to print the docker env info
         with_target_vms(nil, {:single_target=>true}) do |machine|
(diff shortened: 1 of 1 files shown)


navidshaikh:
With OpenShift provided by systemd files, following is the proposed output of subcommand, it exits with status 0 in case if openshift is service is running.

```
vagrant service-manager env openshift
# Set the following environment variables to access the OpenShift
export OPENSHIFT_URL=https://10.1.2.2:8443
export OPENSHIFT_WEB_CONSOLE=https://10.1.2.2:8443/console
# eval "$(vagrant service-manager env openshift)"
```

When open shift service is not running, it exits with status 1 and with following message

```
vagrant service-manager env openshift
OpenShift service is not running in the vagrant box.
```

If user wants to print the status of all the services running inside the box

```
vagrant service-manager env

# Showing the status of providers in the vagrant box:

export DOCKER_HOST=tcp://10.1.2.2:2376
export DOCKER_TLS_VERIFY=1
export DOCKER_MACHINE_NAME=319910c


# Set the following environment variables to access the OpenShift
export OPENSHIFT_URL=https://10.1.2.2:8443
export OPENSHIFT_WEB_CONSOLE=https://10.1.2.2:8443/console
# eval "$(vagrant service-manager env openshift)"

```

bexelbie:
Error messages should probably be # preceded so that users running an $(eval) don't get a bad failure on the eval that would hide the real failure case.

lib/vagrant-service-manager/command.rb
```diff
@@ -73,6 +77,79 @@ def copy_from_box(hIP, hport, husername, hprivate_key_path, source, destination)
...
+        command = "systemctl status openshift"
+        with_target_vms(nil, {:single_target=>true}) do |machine|
+          if machine.communicate.test(command) then
+            # Find the guest IP
+            guest_ip = self.find_machine_ip(machine)
+            openshift_url = "https://#{guest_ip}:#@@OPENSHIFT_PORT"
+            openshift_console_url = "#{openshift_url}/console"
+            self.print_openshift_info(openshift_url, openshift_console_url)
+          else
+            @env.ui.info("# OpenShift service is not running in the vagrant box.")
```
navidshaikh: @bexelbie 

> Error messages should probably be # preceded so that users running an $(eval) don't get a bad failure on the eval that would hide the real failure case.

Guess I pasted the output and figured that before pushing the patch! ;-) Patch has the # before error message.

lib/vagrant-service-manager/command.rb
```diff
@@ -73,6 +77,79 @@ def copy_from_box(hIP, hport, husername, hprivate_key_path, source, destination)
...
+            exit 1
```
navidshaikh: @bexelbie 

> Error messages should probably be # preceded so that users running an $(eval) don't get a bad failure on the eval that would hide the real failure case.

But, the clients consuming the command should first check for the exit status of command before `eval`uation. #45 addresses same.

lib/vagrant-service-manager/command.rb
```diff
@@ -73,6 +77,79 @@ def copy_from_box(hIP, hport, husername, hprivate_key_path, source, destination)
         end
       end
 
+      def check_if_a_service_is_running(machine, service)
```
brgnepal: This looks to be dead code here. Is it being used anywhere?

If it is to be used somewhere in future, then lets check-in this code at that time.

navidshaikh: updated the PR.

lib/vagrant-service-manager/command.rb
```diff
@@ -14,25 +14,29 @@ def self.synopsis
...
         when "env"
             case subcommand
+
             when "docker"
                 self.execute_docker_info
+            when "openshift"
+                self.execute_openshift_info
             when nil
-                self.execute_docker_info
+                # display information about all the providers inside ADB/CDK
```
brgnepal, line 41: Isn't subcommand == nil and `default` case same here?

lib/vagrant-service-manager/command.rb
```diff
@@ -51,10 +55,14 @@ def print_help
 
 Verb:
   env
-    Configures and prints the required environment variables for Docker daemon
-
-Example:
-$vagrant service-manager env docker
+    Display connection information for providers in the box.
+    example:
```
brgnepal: Probably `Example:` looks better here.

navidshaikh: updated.

LalatenduMohanty:
Tested on libvirt + ADB 1.7.0 release condidate

```
[lmohanty@LalatenduM-laptop adb1.7-1]$ vagrant service-manager env openshift
# Set the following environment variables to access the OpenShift
export OPENSHIFT_URL=https://192.168.121.243:8443
export OPENSHIFT_WEB_CONSOLE=https://192.168.121.243:8443/console
# eval "$(vagrant service-manager env openshift)"
```

lib/vagrant-service-manager/command.rb
```diff
@@ -73,29 +81,89 @@ def copy_from_box(hIP, hport, husername, hprivate_key_path, source, destination)
...
+        with_target_vms(nil, {:single_target=>true}) do |machine|
+          # Path to the private_key and where we will store the TLS Certificates
+          secrets_path = File.expand_path("docker", machine.data_dir)
+
+          hIP = machine.ssh_info[:host]
+          hport = machine.ssh_info[:port]
+          husername = machine.ssh_info[:username]
+
+          # Find the guest IP
+          guest_ip = self.find_machine_ip(machine)
```
brgnepal: Probably you forgot to remove argument from `self.find_machine_ip`.

navidshaikh:
Updated the PR based on the discussions for OpenShift display information. Rebased the PR with master.

lib/vagrant-service-manager/command.rb
```diff
@@ -76,29 +84,76 @@ def copy_from_box(hIP, hport, husername, hprivate_key_path, source, destination)
         end
       end
 
-      def execute_docker_info
-        # this execute the operations needed to print the docker env info
+      def check_if_a_service_is_running?(service)
+        command = "systemctl status #{service}"
```
LalatenduMohanty, line 88: @navidshaikh We wont be able to check if k8s is setup and running just using systemctl status #{service}.  So just a info ahead of time. We have to tweak the code to run some other commands to check

navidshaikh, line 88: @LalatenduMohanty : Sure! For k8s, will need to iterate over multiple services and probably do some more. One of the reason I separated out the method to check for service status. Noted!

LalatenduMohanty:
LGTM

PR #47, opened 2016-02-12
Fix #44: Fixes prompt for bringing up machine for help command

### LINKED ISSUE #44: Plugin reports to bring up machine for even help command
Even if the user is issuing `--help` command, plugin prompts user to bring up the machine.
The prompt should be valid for only required commands / subcommands.

navidshaikh:
Fixes: #44 

Exits with status 1 if machine is not up and user has asked for environment variables.

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -11,23 +11,26 @@ module Vagrant
         'provides the IP address:port and tls certificate file location for a docker daemon'
       end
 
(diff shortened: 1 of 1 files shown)


bexelbie:
We should add a separate PR that adds the --help option for commands.  Then this PR should be modified to only be called when we get an actual execution path.

bexelbie:
LGTM

LalatenduMohanty:
LGTM

Issue #12, opened 2016-02-09
service-manager command kills running docker containers on first execution


navidshaikh:
_From @hferentschik on December 16, 2015 12:7_

`vagrant adbinfo` kills all running Docker containers on first execution (for example the running OpenShift container as per https://github.com/redhat-developer-tooling/openshift-vagrant setup).

This is due to the fact that `adbinfo` re-generates TLS certificates for the Docker daemon when it is first called - https://github.com/projectatomic/vagrant-adbinfo/blob/master/lib/command.rb#L65. As part of this the Docker daemon gets restarted which kills all already running containers. In the context of redhat-developer-tooling/openshift-vagrant this means that after a first call to `vagrant adbinfo`, `vagrant provision` needs to be called again, in order to restart the OpenShift container.

The plugin should not do these VM modifications steps as part of the command execution, but it should rather register itself as a provisioner and do the required changes as part of the VM bootstrap.

_Copied from original issue: projectatomic/vagrant-adbinfo#55_

navidshaikh:
_From @hferentschik on December 16, 2015 12:7_

Could someone please add the label foxus:devex.

navidshaikh:
_From @LalatenduMohanty on January 27, 2016 15:11_

@navidshaikh @hferentschik Is it already resolved for CDK?

navidshaikh:
_From @hferentschik on January 28, 2016 11:31_

> Is it already resolved for CDK?

No, we are still not using OpenShift as a service, since there are still issues with the provided scripts. And even then, switching to OpenShift as a service might help in this particular case, the true solution is in the implementation of this plugin.

navidshaikh:
_From @hferentschik on January 28, 2016 11:52_

I also quickly tested with the service scripts. In this case the OpenShift service indeed restarts after the Docker service restart, so OpenShift comes back up again. However, this is not a proper solution. Other containers which might have been started are still "lost" and one has a considerable downtime of the OpenShift service.

Issue #21, opened 2016-02-09
Generate TLS certs during provisioning


navidshaikh:
_From @bexelbie on November 22, 2015 12:37_

Plugin should run automatically when the ADB is started like https://github.com/projectatomic/adb-vagrant-registration/tree/master/lib/vagrant-registration

@strzibny can you provide any guidance?

_Copied from original issue: projectatomic/vagrant-adbinfo#34_

navidshaikh:
_From @bexelbie on November 22, 2015 21:12_

@LalatenduMohanty how does the prompting for registration for the RHEL CDK work? could we adapt that?

navidshaikh:
_From @strzibny on November 23, 2015 10:27_

@bexelbie What would be the use case? You want automatic way of exposing the env variables or? Something I suggested here https://github.com/projectatomic/vagrant-adbinfo/issues/13 but as part of vagrant up process or?

Of course technically speaking I can help.

navidshaikh:
_From @LalatenduMohanty on November 23, 2015 10:37_

@bexelbie RHEL CDK expect the user to install the plugin. I does not happen automatically. I think we need to sync with the team which is working on a installer for CDK.

navidshaikh:
_From @bexelbie on November 23, 2015 11:20_

@strzibny @strizbny I'd like to see vagrant adbinfo run automatically after vagrant up. It seems that vagrant register is called automatically as part of vagrant up, somehow. 

Re #13 can you provide some advice in the best way to export those variables automatically from Python? My foo there is not as strong as it should be or I need more coffee and sleep.

navidshaikh:
_From @bexelbie on November 23, 2015 11:20_

@LalatenduMohanty @LalatenduMohanty I meant auto-run after vagrant up, not auto-install.

navidshaikh:
_From @LalatenduMohanty on November 23, 2015 11:27_

@bexelbie ah, I missed that. Yes, if you have vagrant-registration installed and it is a RHEL guest ,  its asks for username and password for registration every time you do a "vagrant up" . 

I think its good idea to put auto detection (guest capability) code in the plugin. 
e.g  https://github.com/projectatomic/adb-vagrant-registration/blob/master/plugins/guests/redhat/plugin.rb

navidshaikh:
_From @bexelbie on November 23, 2015 12:15_

[...]

Issue #49, opened 2016-02-15
Code refactoring


brgnepal:
After going through the plugin's code a while, I think we need to do major refactoring to meet the [Ruby language's style and standards](https://github.com/airbnb/ruby).  

I will be working on it after some stable release version.

Action items:
- Follow Ruby style guide throughout the plugin
- Autoload files in `vagrant-service-manager.rb` 
- Change plugin names from `Vagrant` to `VagrantPlugins`
- YML approach for doc strings(help text)
- #79 (Service driven command structure)

budhrg:
Expect heavy refactoring post `v0.0.2`.

hferentschik:
+1 Absolutely required. It should also take into account that the plugin in the future does not only provide commands, but also uses other Vagrant plugin mechanisms (for example action hooks).

brgnepal:
@hferentschik : Action hook has been used for configuring providers like `docker, openshift, ...` on `vagrant up`. PR here as #54 .  It would be available in next release.

bexelbie:
Pending #54 I also think that the code in `command.rb` should be refactored so that the service specific bits live somewhere like `lib/vagrant-service-manager/services/docker.rb`.  Other code like find_machine_ip can go in a common library.

If this isn't part of this PR (and I can go both ways) we need to do it ASAP.

brgnepal:
@bexelbie lets keep it in this PR only. Will focus on refactoring after next release.

bexelbie:
Moved my last comment to #135

PR #54, opened 2016-02-17
Config + hook approach for restarting docker

brgnepal:
Fix: #12 and #21.

With this approach `Vagrantfile` will look like

```
 Vagrant.configure(2) do |config|
  config.vm.box = "projectatomic/adb"
  config.servicemanager.providers = 'docker'
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.0.3 Feb 25, 2016
+- Fix #12 and #21: Restart docker service on 'vagrant up' @budhrg
 - Fix #74: vagrant-service-manager plugin version 0.0.3 release @navidshaikh
 - Fix #45: Adds exit status for commands and invalid commands @navidshaikh
 - Enhanced the developer instructions for developing the plugin in README @budhrg
--- a/README.md
+++ b/README.md
@@ -1,13 +1,6 @@
 # vagrant-service-manager
 
-Provides setup information, including environment
-variables and certificates, required to access
-services provided by an [Atomic Developer Bundle
-(ADB)](https://github.com/projectatomic/adb-atomic-developer-bundle).
-This plugin makes it easier to use the ADB with host-based tools
-such as Eclipse and the docker and kubernetes CLI commands.
-Details on this usage pattern can be found in the [ADB
-Documentation](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/using.rst).
+This plugin provides setup information, including environment variables and certificates, required to access services provided by an [Atomic Developer Bundle (ADB)](https://github.com/projectatomic/adb-atomic-developer-bundle).  This plugin makes it easier to use the ADB with host-based tools such as Eclipse and the docker and kubernetes CLI commands.  Details on this usage pattern can be found in the [ADB Documentation](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/using.rst).
 
 ##Objective
 
@@ -27,16 +20,28 @@ producing complex, multi-container applications.
 
 ## Quick Start
 
-1. Install `vagrant-service-manager` plugin
+1. Install `vagrant-service-manager` plugin:
 
         vagrant plugin install vagrant-service-manager
 
-2. Get the [Vagrantfile](Vagrantfile)
-and start the ADB using `vagrant up`.
-[More](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/installing.rst)
-documentation on setting up ADB.
+2. Get a Vagrantfile for your box. Users of the
+[Atomic Developer Bundle (ADB)](https://github.com/projectatomic/adb-atomic-developer-bundle) should download a [Vagrantfile from the repository](https://github.com/projectatomic/adb-atomic-developer-bundle/tree/master/components).
 
-3. Run the plugin to get environment variables and certificates
+3. Enable your desired service(s) in [Vagrantfile](Vagrantfile) as:
+
+        config.servicemanager.services = 'openshift'
+
+   *Note*
+
+   * `docker` is default service and does not require above configuration.
+   * Enable multiple services as comma separated list. Eg: 'docker, openshift'
+
+5. Start the ADB using `vagrant up`. Users of the ADB may wish to consult the
+[Installation Documentation](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/installing.rst).
+
+6. Run the plugin to get environment variables and certificates:
+
+        $ vagrant service-manager env docker
 
         # Copying TLS certificates to /home/nshaikh/vagrant/adb1.7/.vagrant/machines/default/virtualbox/docker
         # Set the following environment variables to enable access to the
@@ -48,7 +53,7 @@ documentation on setting up ADB.
         # run following command to configure your shell:
         # eval "$(vagrant service-manager env docker)"
 
-4. Begin using your host-based tools.
+7. Begin using your host-based tools.
 
 ## Exit codes
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,9 +1,23 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+unless Vagrant.has_plugin?("vagrant-service-manager")
+  $stderr.puts <<-MSG
+    vagrant-service-manager plugin is required for projectatomic/adb.
+    Kindly install the plugin: `$ vagrant plugin install vagrant-service-manager`
+  MSG
+  exit 126
+end
+
 Vagrant.configure(2) do |config|
+
   config.vm.box = "projectatomic/adb"
 
   config.vm.network "private_network", type: "dhcp"
 
+  # This is the default setup
+  # config.servicemanager.services = 'docker'
+
+  # Enable multiple services as comma separated list.
+  # config.servicemanager.services = 'docker, openshift'
 end
--- a/lib/vagrant-service-manager/config.rb
+++ b/lib/vagrant-service-manager/config.rb
@@ -1,0 +1,50 @@
+require 'set'
+
+module Vagrant
+  module ServiceManager
+    SERVICES = ['docker', 'openshift']
+
+    class Config < Vagrant.plugin('2', :config)
+      attr_accessor :services
+
+      DEFAULTS = {
+        services: 'docker'
+      }
+
+      def initialize
+        super
+        @services = UNSET_VALUE
+      end
+
+      def finalize!
+        DEFAULTS.each do |name, value|
+          if instance_variable_get('@' + name.to_s) == UNSET_VALUE
+            instance_variable_set '@' + name.to_s, value
+          end
+        end
+      end
+
+      def validate(machine)
+        errors = _detected_errors
+        errors.concat(validate_services)
+        { "servicemanager" => errors }
+      end
+
+      private
+
+      def validate_services
+        errors = []
+
+        unless is_supported_services?
+          errors << "services should be subset of #{SERVICES.inspect}.}"
+        end
+
+        errors
+      end
+
+      def is_supported_services?
+        @services.split(',').map(&:strip).to_set.subset?(SERVICES.to_set)
+      end
+    end
+  end
+end
--- a/lib/vagrant-service-manager/plugin.rb
+++ b/lib/vagrant-service-manager/plugin.rb
@@ -1,3 +1,6 @@
+# Loads all services
+Dir["#{File.dirname(__FILE__)}/services/*.rb"].each { |f| require_relative f }
+
 module Vagrant
   module ServiceManager
     class Plugin < Vagrant.plugin('2')
@@ -7,6 +10,15 @@ module Vagrant
       command 'service-manager' do
         require_relative 'command'
         Command
+      end
+
+      config 'servicemanager' do
+        require_relative 'config'
+        Config
+      end
+
+      action_hook(:servicemanager, :machine_action_up) do |hook|
+        hook.append(Service::Docker)
       end
     end
   end
--- a/lib/vagrant-service-manager/services/docker.rb
+++ b/lib/vagrant-service-manager/services/docker.rb
@@ -1,0 +1,29 @@
+module Vagrant
+  module ServiceManager
+    SUPPORTED_BOXES = ['adb', 'cdk']
+
+    module Service
+      class Docker
+        def initialize(app, env)
+          @app = app
+          @machine = env[:machine]
+          @ui = env[:ui]
+        end
+
+        def call(env)
+          if SUPPORTED_BOXES.include? @machine.guest.capability(:flavor)
+            command = "sudo rm /etc/docker/ca.pem && sudo systemctl restart docker"
+            @machine.communicate.execute(command) do |type, data|
+              if type == :stderr
+                @ui.error(data)
+                exit 126
+              end
+            end
+          end
+
+          @app.call(env)
+        end
+      end
+    end
+  end
+end
--- a/lib/vagrant-service-manager.rb
+++ b/lib/vagrant-service-manager.rb
@@ -9,10 +9,13 @@ require 'vagrant-service-manager/command'
 require 'vagrant-service-manager/os'
 
 module Vagrant
-  module DockerInfo
+  module ServiceManager
     # Returns the path to the source of this plugin
     def self.source_root
       @source_root ||= Pathname.new(File.expand_path('../../', __FILE__))
     end
+
+    # Temporally load the extra capabilities files for Red Hat
+    load(File.join(source_root, 'plugins/guests/redhat/plugin.rb'))
   end
 end
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -192,13 +192,6 @@ Verb:
           # First, get the TLS Certificates, if needed
           if !File.directory?(secrets_path) then
 
-            # Regenerate the certs and restart docker daemon in case of the new ADB box and for VirtualBox provider
-            if machine.provider_name == :virtualbox then
-              # `test` checks if the file exists, and then regenerates the certs and restart the docker daemon, else do nothing.
-              command2 = "test ! -f /opt/adb/cert-gen.sh || (sudo rm /etc/docker/ca.pem && sudo systemctl restart docker)"
-              machine.communicate.execute(command2)
-            end
-
             # Get the private key
             hprivate_key_path = machine.ssh_info[:private_key_path][0]
 
--- a/plugins/guests/redhat/cap/flavor.rb
+++ b/plugins/guests/redhat/cap/flavor.rb
@@ -1,0 +1,17 @@
+module VagrantPlugins
+  module GuestRedHat
+    module Cap
+      class Flavor
+        def self.flavor(machine)
+          machine.communicate.sudo("grep VARIANT_ID /etc/os-release") do |type, data|
+            if type == :stderr
+              @env.ui.error(data)
+              exit 126
+            end
+            return data.chomp.gsub(/"/, '').split("=").last
+          end
+        end
+      end
+    end
+  end
+end
--- a/plugins/guests/redhat/plugin.rb
+++ b/plugins/guests/redhat/plugin.rb
@@ -1,0 +1,12 @@
+require 'vagrant'
+
+module VagrantPlugins
+  module GuestRedHat
+    class Plugin < Vagrant.plugin('2')
+      guest_capability('redhat', 'flavor') do
+        require_relative 'cap/flavor'
+        Cap::Flavor
+      end
+    end
+  end
+end


README.md
```diff
@@ -12,13 +12,32 @@ The [Atomic Developer Bundle](https://github.com/projectatomic/adb-atomic-develo
...
+2. Get the [Vagrantfile](Vagrantfile).
 
-3. Run the plugin to get environment variables and certificates
+3. Enable your desired provider(s) in [Vagrantfile](Vagrantfile) as:
+  ```
+    Vagrant.configure(2) do |config|
+
+      config.vm.box = "projectatomic/adb"
+
+      config.vm.network "private_network", type: "dhcp"
```
bexelbie: This example should probably be the only line cited.

The Vagrantfile should contain comments with the list and format.

lib/vagrant-service-manager/action/restart_service.rb
```diff
@@ -0,0 +1,34 @@
...
+              @ui.error(data)
+            else
+              @ui.info("# Restarted #{provider} service.")
+            end
+          end
+        end
+
+        def call(env)
+          @providers.each do |provider|
+            command = "sudo systemctl restart #{provider}"
```
bexelbie: I feel like each provider should have it's own module that defines restart for it.  This way we don't have to worry about someone passing in unexpected values.  i.e. vagrant service-manager restart httpd - if we dont' have apache defined.

This also may make it cleaner to add new providers down the road.  

for POC this is OK, but not for merge.

lib/vagrant-service-manager/config.rb
```diff
@@ -0,0 +1,22 @@
...
+      attr_accessor :providers
+
+      def initialize
+        super
+        @providers = UNSET_VALUE
+      end
+
+      def finalize!
+        # enable docker as provider by default
+        @providers = 'docker' if @providers == UNSET_VALUE
```
bexelbie: This default needs to be documented.

bexelbie:
Can we also get a changelog modification?  Can you build a test ADB and gem that we can send out for use as a "nightly"?

lib/vagrant-service-manager/config.rb
brgnepal: Done and updated in README.

lib/vagrant-service-manager/action/restart_service.rb
navidshaikh: > This way we don't have to worry about someone passing in unexpected values. i.e. vagrant service-manager restart httpd - if we dont' have apache defined.

+1

LalatenduMohanty:
We should squash the commits in to one.

lib/vagrant-service-manager/action/restart_service.rb
brgnepal: @bexelbie 

> worry about someone passing in unexpected values. i.e. vagrant service-manager restart httpd - if we dont' have apache defined.

Since the `config + hook` code will be executed on `vagrant up` which is based on Vagrantfile configuration for `providers(config.servicemanager.providers)`, we need not to worry about someone passing unexpected values as the values will be fixed(like `docker, openshift, kubernetes` etc). 
However, we can also validate the providers list passed in `Vagrantfile` and throw error if wrong provider is passed.

bexelbie: The benefit of a more modular approach, which I believe is used by atomic app too, is that it is easier for those unfamiliar with the whole code base to add a new service.

Vagrantfile
```diff
@@ -1,9 +1,20 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+unless Vagrant.has_plugin?("vagrant-service-manager")
+  raise <<-MSG
```
bexelbie: Can you examine this code relative to this: https://github.com/projectatomic/adb-atomic-developer-bundle/issues/245

Can we do a cleaner error and exit?

brgnepal:
@LalatenduMohanty done with sqashing commit.

brgnepal:
@bexelbie,

Latest changes includes provider centric service operation.

lib/vagrant-service-manager/config.rb
```diff
@@ -0,0 +1,39 @@
...
+        errors.concat(validate_providers)
+        { "servicemanager" => errors }
+      end
+
+      private
+
+      def validate_providers
+        errors = []
+
+        if !Set[@providers.split(',').map(&:strip)].subset?(Set[PROVIDERS])
```
brgnepal: If specified provider(s) in configuration is not listed in `PROVIDERS` then it errors out.

Vagrantfile
```diff
@@ -1,9 +1,20 @@
...
+end
+
 Vagrant.configure(2) do |config|
+
   config.vm.box = "projectatomic/adb"
 
   config.vm.network "private_network", type: "dhcp"
 
+  # enable multiple providers as comma separated list. Eg: 'docker, openshift'
+  config.servicemanager.providers = 'docker'
```
hferentschik: why the need for explicitly enabling this? Does this also mean that the docker certificates are only generated if the docker provider is set? They should really be generated regardless. 

The aim is zero conf Vagrantfile. I think this config should not be required.

navidshaikh: > Does this also mean that the docker certificates are only generated if the docker provider is set? They should really be generated regardless.

I agree.

> The aim is zero conf Vagrantfile. I think this config should not be required.

For `docker` provider yes, for other providers we may need config in Vagrantfile.

hferentschik: > For docker provider yes, for other providers we may need config in Vagrantfile.

Ok, seems reasonable. We seem to agree the proper configuration of Docker is a given. There is no need to explicitly configure it. What other providers are there? Kubernetes vs OpenShift? Is really OpenShift not a super-set of Kubernetes here? Also OpenShift is the default. It could be generated unless something else is explicitly configured. Again, the aim is zero config _Vagrantfile_. Let's discuss the need for configuring other providers, but Docker should just work.

navidshaikh: > What other providers are there?

Apart from kube and openshift, Mesos-Marathon

navidshaikh: > Does this also mean that the docker certificates are only generated if the docker provider is set? They should really be generated regardless.

Need to check if the approach we are following (config + hook) can trigger task of configuring-docker without any mention of config in Vagrantfile explicitly.

README.md
```diff
@@ -12,13 +12,23 @@ The [Atomic Developer Bundle](https://github.com/projectatomic/adb-atomic-develo
...
 
         vagrant plugin install vagrant-service-manager
 
-2. Get the [Vagrantfile](Vagrantfile) and start the ADB using `vagrant up`. [More](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/installing.rst) documentation on setting up ADB.
+2. Get the [Vagrantfile](Vagrantfile).
 
-3. Run the plugin to get environment variables and certificates
+3. Enable your desired provider(s) in [Vagrantfile](Vagrantfile) as:
+
+        config.servicemanager.providers = 'docker'
```
bexelbie: we should show an example of starting multiple services

Also, even though these are container providers, perhaps we should call them services or use the word provision i.e.

config.servicemanager.services = 'docker'
or
config.servicemanager.provision = 'docker'

bexelbie:
Should execute_docker_info in lib/vagrant-service-manager/command.rb actually be in lib/vagrant-service-manager/providers/docker.rb ?

lib/vagrant-service-manager/plugin.rb
```diff
@@ -1,3 +1,5 @@
+require_relative 'providers/docker'
```
bexelbie: Can we require_relative 'providers/*' to eliminate the need to edit this file when adding new services?

bexelbie: Also, should this be called services/ ?

bexelbie:
Can we somehow validate the commands in vagrant-service-manager/command.rb by using what is provided in the ruby file in providers?

Vagrantfile
brgnepal: > Does this also mean that the docker certificates are only generated if the docker provider is set? They should really be generated regardless.

@navidshaikh , @hferentschik 
It is working. :smile: .
Just omit config line from Vagrantfile. [This line](https://github.com/budhrg/vagrant-service-manager/blob/config-hook/lib/vagrant-service-manager/config.rb#L17) is doing the trick for it.

README.md
brgnepal: @navidshaikh , @LalatenduMohanty any opinion?
:+1: for `service`.

lib/vagrant-service-manager/plugin.rb
brgnepal: @bexelbie , @navidshaikh : 
Would it better if we use consistent naming? I would like to agree with the term `service`.
This line will be `require_relative 'services/docker'` or 
`Dir.glob('lib/*.rb') { |f| require_relative f }`(best I could) for automatically adding new services. 

Also, `Vagrantfile` line as:
`config.servicemanager.services = 'openshift'`([docker default hence this line will not be needed](https://goo.gl/0F6ttj)).

And class name `Provider` to `Service`.

budhrg:
> Can we somehow validate the commands in vagrant-service-manager/command.rb by using what is provided in the ruby file in providers?

@bexelbie : [`machine`](https://goo.gl/tpmC9Z) object will have those configurations and could be accessed as `machine.config.servicemanager.providers`.

brgnepal:
> Should execute_docker_info in lib/vagrant-service-manager/command.rb actually be in lib/vagrant-service-manager/providers/docker.rb ?

@bexelbie , @LalatenduMohanty , @navidshaikh : Let's discuss such changes as part of #49 .  There are other more things I need to discuss as part of refactoring.

Vagrantfile
navidshaikh: @budhrg 

> It is working. :smile: .
> Just omit config line from Vagrantfile. This line is doing the trick for it.

Excellent!

bexelbie:
@budhrg Can you think about this idea, even if it comes in a separate patch.

If https://github.com/projectatomic/adb-utils/pull/45 merges, it would be nice to be able to do this in the vagrant file

```
config.vm.service-manager.openshift.image_name = "openshift/origin"
config.vm.service-manager.openshift.image_tag = "latest"
config.vm.service-manager.openshift.docker_registry = "docker.io"

And have those values available for a later provisioner shell call.

bonus points if the presence of any those caused to assume this line:

config.vm.service-manager.provision = 'openshift'
```

hferentschik:
Awesome. Now we start talking. This is exactly what I had in mind. The service-manager basically offering nice and simple extension/config options to handle some of these default things. Much better than fiddling around with _sed_ and co in some shell provisioning script. It would be awesome, if we could get this all sorted for GA. Also needs of course be well documented.

brgnepal:
@bexelbie , @hferentschik : 

> config.vm.service-manager.openshift.image_name = "openshift/origin"
> config.vm.service-manager.openshift.image_tag = "latest"
> config.vm.service-manager.openshift.docker_registry = "docker.io"
> ....... or more such lines possibly ....

How about this?

```
config.servicemanager.provision(or service) = 'openshift'   # enable openshift
config.servicemanager.service.openshift = { image: "openshift/origin", tag: "latest", docker_registry: "docker.io", (other values too...)}
# above values will be available to later provisioners
```

Above lines are just imaginary which I think possible and would be reference for my working.
Please correct me if I interpreted it differently.

bexelbie:
@budhrg I'm not a vagrant expert but it feels like making the user write json in the Vagrantfile is weird and an anti-pattern. The other feels cleaner to me. Why are you suggesting this?

brgnepal:
@bexelbie Oh ok. I thought it would be only one line than having multiple lines. 
Also agree that multi-line approach would be far more comfortable for users than `json` which would be syntax error prone.
Thanks for clarification.

hferentschik:
> Also agree that multi-line approach would be far more comfortable for users than json 

+1 We basically offer one option for everything we think is relevant and should be configurable.

brgnepal:
@bexelbie 

> Can we somehow validate the commands in vagrant-service-manager/command.rb by using what is provided in the ruby file in providers?

Need more info about it. Could you elaborate it little bit?

LalatenduMohanty:
> Let's discuss such changes as part of #49 . There are other more things I need to discuss as part of refactoring.

Agree. We might want to do that in a separate PR

LalatenduMohanty:
> @budhrg I'm not a vagrant expert but it feels like making the user write json in the Vagrantfile is weird and an anti-pattern.

+1. We dont need json kind of structure

lib/vagrant-service-manager/plugin.rb
bexelbie: +1 service(s) - it really opens up the options for this plugin and makes the language so much better

bexelbie:
@budhrg What I mean here is that it would be nice if the service could define what commands it could interact with. So for example if the foo service doesn't have an env verb then 'env foo' fails but we don't have to create a command table and order to know this. I suspect the word introspection somehow comes into play here.

lib/vagrant-service-manager/plugin.rb
hferentschik: +1 Service is the better term. I think we need to aim for a consistent terminology not only in this plugin, but across the whole ADB/CDK. "The CDK offers a OpenShift service" or "To enable the Kubernetes service ...", etc

brgnepal:
+1 for the idea :smile: .
@bexelbie - This could be possible with Ruby Meta-programming(introspection) as what I could infer from your comment. Let's discuss it in the issue #79(`Service driven command structure`) as it has to be implemented well as it includes metaprogramming.*

> - Power comes with great responsibility

brgnepal:
@bexelbie , @LalatenduMohanty , @navidshaikh - Done with suggested changes from my side.

README.md
```diff
@@ -12,13 +12,23 @@ The [Atomic Developer Bundle](https://github.com/projectatomic/adb-atomic-develo
 
 ## Quick Start
 
-1. Install `vagrant-service-manager` plugin
+1. Install `vagrant-service-manager` plugin:
 
         vagrant plugin install vagrant-service-manager
 
-2. Get the [Vagrantfile](Vagrantfile) and start the ADB using `vagrant up`. [More](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/installing.rst) documentation on setting up ADB.
+2. Get the [Vagrantfile](Vagrantfile).
```
bexelbie: This needs to be a reference to the ADB repo.  The Vagrantfile in this repo is for testing and should possibly be deleted or augmented.

I suggest:
1. Get a Vagrantfile for your box.  Users of the [Atomic Developer Bundle (ADB)](https://github.com/projectatomic/adb-atomic-developer-bundle/) should download a [Vagrantfile from the repository.](https://github.com/projectatomic/adb-atomic-developer-bundle/tree/master/components)

README.md
```diff
@@ -12,13 +12,23 @@ The [Atomic Developer Bundle](https://github.com/projectatomic/adb-atomic-develo
...
-3. Run the plugin to get environment variables and certificates
+3. Enable your desired service(s) in [Vagrantfile](Vagrantfile) as:
+
+        config.servicemanager.services = 'openshift'
+
+   *Note*
+
+   `docker` is default service and does not require above configuration.
+
+4. Start the ADB using `vagrant up`. [More](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/installing.rst) documentation on setting up ADB.
```
bexelbie: Suggestion:
1. Start the ADB using `vagrant up`. Users of the ADB may wish to consult the [Installation Documentation](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/installing.rst).

README.md
```diff
@@ -12,13 +12,23 @@ The [Atomic Developer Bundle](https://github.com/projectatomic/adb-atomic-develo
...
+
+5. Run the plugin to get environment variables and certificates:
 
         # Copying TLS certificates to /home/nshaikh/vagrant/adb1.7/.vagrant/machines/default/virtualbox/docker
```
bexelbie, line 46: Let's update with newer output that shows the full command.  Perhaps we should do a `vagrant service-manager env` on a box with openshift running.

Vagrantfile
bexelbie: I'd still like to see this as a print to STDERR and exit combination.

README.md
```diff
@@ -12,13 +12,23 @@ The [Atomic Developer Bundle](https://github.com/projectatomic/adb-atomic-develo
...
 
         vagrant plugin install vagrant-service-manager
 
-2. Get the [Vagrantfile](Vagrantfile) and start the ADB using `vagrant up`. [More](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/installing.rst) documentation on setting up ADB.
+2. Get the [Vagrantfile](Vagrantfile).
 
-3. Run the plugin to get environment variables and certificates
+3. Enable your desired service(s) in [Vagrantfile](Vagrantfile) as:
+
+        config.servicemanager.services = 'openshift'
```
bexelbie, line 32: This should be "service-manager" to match the plugin name, imho.

lib/vagrant-service-manager/plugin.rb
```diff
@@ -8,6 +11,15 @@ class Plugin < Vagrant.plugin('2')
         require_relative 'command'
         Command
       end
+
+      config 'servicemanager' do
+        require_relative 'config'
+        Config
+      end
+
+      action_hook(:servicemanager, :machine_action_up) do |hook|
```
bexelbie, line 20: This should only run when a config.service-manager line is in the Vagrantfile, otherwise it runs on every box I start whether it should or not.  The only other way is to detect ADB and CDK which we can't do afaik right now.

Vagrantfile
```diff
@@ -1,9 +1,19 @@
...
+  MSG
+end
+
 Vagrant.configure(2) do |config|
+
   config.vm.box = "projectatomic/adb"
 
   config.vm.network "private_network", type: "dhcp"
 
+  # Enable multiple services as comma separated list
```
bexelbie: Can we add:

```
# This is the default setup
# config.service-manager.service = 'docker'
```

README.md
brgnepal, line 32: @bexelbie : Here, it is not accepting `-` based name.

lib/vagrant-service-manager/plugin.rb
budhrg: Ya @bexelbie . We can omit following lines:

```
config 'servicemanager' do
       require_relative 'config'
       Config
end
```
and respective `config.rb` file since it is not doing much ATM.

bexelbie:
I added some line notes.

Vagrantfile
brgnepal: About enabling multiple services?
Should we add it in `README`?

hferentschik: > Should we add it in README

+1

navidshaikh:
If we are looking for finding out the box OS, in case of non supported boxes, we should just return silently without any trace on the normal execution flow of `vagrant up`.

This should help user have the plugin installed and not having any interception while working with non-supported boxes.

brgnepal:
@LalatenduMohanty , @navidshaikh , @bexelbie updated with changes. Now only `adb` and `cdk` boxes will run this plugin.

Also tested with "cdk" with box `rhel-cdk-kubernetes-7.2-14`.

lib/vagrant-service-manager/services/docker.rb
```diff
@@ -0,0 +1,29 @@
...
+          @ui = env[:ui]
+        end
+
+        def call(env)
+          if SUPPORTED_BOXES.include? @machine.guest.capability(:flavor)
+            command = "sudo rm /etc/docker/ca.pem && sudo systemctl restart docker"
+            @machine.communicate.execute(command) do |type, data|
+              if type == :stderr
+                @ui.error(data)
+                exit 126
```
navidshaikh, line 19: Good!

navidshaikh:
@LalatenduMohanty : Once you have tested the PR against non supported box, can you give feedback ?

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
+- Fix #12 and #21: Restart docker service on 'vagrant up' @budhrg
```
navidshaikh: Kindly put this changelog on v0.0.3 release

navidshaikh:
I tested this PR against projectatomic/adb and fedora/23-cloud-base for ensuring it identifies between supported and non supported boxes, works fine for me.

lib/vagrant-service-manager/plugin.rb
navidshaikh, line 20: After adding the guest detection capability, this is resolved. Now, it will only run against supported [ADB / CDK] boxes and `docker` is the default provider.

navidshaikh:
LGTM!

lib/vagrant-service-manager/services/docker.rb
```diff
@@ -0,0 +1,29 @@
...
+    module Service
+      class Docker
+        def initialize(app, env)
+          @app = app
+          @machine = env[:machine]
+          @ui = env[:ui]
+        end
+
+        def call(env)
+          if SUPPORTED_BOXES.include? @machine.guest.capability(:flavor)
```
bexelbie, line 14: Why is this inside of the docker hook and not at a higher level?  This structure means that every time we add a new service we have to write this same code again.

I am going to LGTM this, but I am opening an issue to get this fixed as we need to get openshift, etc. added.

bexelbie:
LGTM with new issue coming

lib/vagrant-service-manager/services/docker.rb
brgnepal, line 14: Agreed @bexelbie . Will work on it.

PR #61, opened 2016-02-18
Fix #8: Adds subcommand for printing box version

### LINKED ISSUE #8: Plugin should report CDK version
_From @hferentschik on January 6, 2016 20:38_

Something like `vagrant adb version` (this ties also into renaming the plugin and exposing a whole set of sub-commands).

I am not sure whether it is right now even possible to inspect anything in the VM to determine the version. @LalatenduMohanty, @praveenkumar !?

_Copied from original issue: projectatomic/vagrant-adbinfo#64_

navidshaikh:
Fix #8 

```
vagrant service-manager box version
Container Development Kit (CDK) 2.0.beta4
```

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -43,6 +43,17 @@ module Vagrant
             else
                 self.print_help
             end
+
+        when "box"
+          self.exit_if_machine_not_running
+          case subcommand
+
+          when "version"
+            self.print_vagrant_box_version
+          else
+            self.print_help
+          end
+
         when "help"
             self.print_help
         else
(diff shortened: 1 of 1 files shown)


lib/vagrant-service-manager/command.rb
```diff
@@ -217,6 +228,28 @@ def print_docker_env_info(guest_ip, port, secrets_path, machine_uuid)
           @env.ui.info(puts(message))
         end
       end
+
+      def print_vagrant_box_version
+        # Prints the version of the vagrant box, parses /etc/os-release for version
+        @@OS_RELEASE = "/etc/os-release"
```
brgnepal: Why as class variable?

As this is constant, `OS_RELEASE` would be fine. 

However, we should have a separate file as module to store all constants. Let's keep that as part of refactoring #49 .

navidshaikh: > As this is constant, OS_RELEASE would be fine.

ACK! update coming..

> However, we should have a separate file as module to store all constants. Let's keep that as part of refactoring #49 .

+1

lib/vagrant-service-manager/command.rb
```diff
@@ -43,6 +43,17 @@ def execute
             else
                 self.print_help
             end
+
+        when "box"
+          self.exit_if_machine_not_running
+          case subcommand
+
```
brgnepal: Unnecessary empty line here at line 50.

LalatenduMohanty:
LGTM

PR #62, opened 2016-02-18
Fix #45: Adds exit status for commands and invalid commands

### LINKED ISSUE #45: Define and implement the exit code / status for commands exectuion
Since the output of plugin could be consumed by other tooling, different exit statuses from the plugin commands and sub-commands need to be defined based on the command execution and their statuses.

This should help other tools or user know if the execution of command was successful without parsing the output of command.

navidshaikh:
Fix #45 
- Exit with status 1, if:
  - Wrong command given / Catchall for general errors
- Exit with status 3 if:
  - Machine is not running
- Exit with status 126, if
  - Service inside the box is not running and user has asked for its connection information Or Command invoked cannot execute

Reference: http://www.tldp.org/LDP/abs/html/exitcodes.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Changelog
 
-## Unreleased
-- Enhanced the developer instructions for developing the plugin in README
+## v0.0.3 Feb 25, 2016
+-
+- Enhanced the developer instructions for developing the plugin in README @budhrg
+- Fix #45: Adds exit status for commands and invalid commands @navidshaikh
 
 ## v0.0.2 Feb 17, 2016
 - Fixes #53: Prep for version v0.0.2
--- a/README.md
+++ b/README.md
@@ -32,6 +32,18 @@ The [Atomic Developer Bundle](https://github.com/projectatomic/adb-atomic-develo
 
 4. Begin using your host-based tools.
 
+## Exit codes
+
+The following table lists the plugin's exit codes and their meaning:
+
+Exit Code Number | Meaning
+---------------  |-------------------------------------------------------------------------
+0                | No error
+1                | Catch all for general errors / Wrong sub-command or option given
+3                | Vagrant box is not running and must be before this command can succeed
+126              | A service inside the box is not running / Command invoked cannot execute
+
+
 ## Get Involved/Contact Us
 
   * IRC: #atomic and #nulecule on freenode
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -12,7 +12,7 @@ module Vagrant
       end
 
       def exit_if_machine_not_running
-        # Exit from plugin with status 1 if machine is not running
+        # Exit from plugin with status 3 if machine is not running
         with_target_vms(nil, {:single_target=>true}) do |machine|
           if machine.state.id != :running then
             message = <<-eos
@@ -21,7 +21,7 @@ module Vagrant
   vagrant up
               eos
             @env.ui.error(message)
-            exit 1
+            exit 3
           end
         end
       end
@@ -50,7 +50,7 @@ module Vagrant
             # display information about all the providers inside ADB/CDK
             self.print_all_provider_info
           else
-            self.print_help
+            self.print_help(1)
           end
         when "box"
           self.exit_if_machine_not_running
@@ -62,7 +62,7 @@ module Vagrant
             when "--script-readable"
               self.print_vagrant_box_version(true)
             else
-                self.print_help
+                self.print_help(1)
             end
           else
             self.print_help
@@ -70,11 +70,11 @@ module Vagrant
         when "help"
             self.print_help
         else
-            self.print_help
+            self.print_help(1)
         end
       end
 
-      def print_help
+      def print_help(exit_status=0)
         help_text = <<-help
 Service manager for services inside vagrant box.
 
@@ -97,6 +97,7 @@ Verb:
           --script-readable : Display the version and release in script readable (key=value) form
         help
         @env.ui.info(help_text)
+        exit exit_status
       end
 
       def copy_from_box(hIP, hport, husername, hprivate_key_path, source, destination)
@@ -138,8 +139,8 @@ Verb:
           openshift_console_url = "#{openshift_url}/console"
           self.print_openshift_info(openshift_url, openshift_console_url)
         else
-          @env.ui.info("# OpenShift service is not running in the vagrant box.")
-          exit 1
+          @env.ui.error("# OpenShift service is not running in the vagrant box.")
+          exit 126
         end
       end
 
@@ -255,7 +256,7 @@ setx DOCKER_MACHINE_NAME #{machine_uuid[0..6]}
           machine.communicate.execute(command) do |type, data|
             if type == :stderr
               @env.ui.error(data)
-              exit 2
+              exit 126
             end
 
             if !script_readable


brgnepal:
+1 for adding argument to `print_help` to accept status.

LGTM

LalatenduMohanty:
@navidshaikh here is the guideline for exit codes http://www.tldp.org/LDP/abs/html/exitcodes.html 
I think we should exit 1 or 126  when machine is not running
Let me know your thoughts.

navidshaikh:
@LalatenduMohanty : 

> I think we should exit 1 or 126 when machine is not running
> Let me know your thoughts.

I think wrong command given is a general error, and we are catching it by exiting with 1.
However  "machine is not running" is something specific to plugin and use case.

LalatenduMohanty:
@navidshaikh I dont have a strong counter argument as I am not sure what will be the right exit code here i.e. when OpenShift service is not running . So unless @bexelbie has some opinion , we are good to go.

navidshaikh:
PR needs rebase and update with exit statuses for merged PRs.

lib/vagrant-service-manager/command.rb
```diff
@@ -21,7 +21,7 @@ def exit_if_machine_not_running
   vagrant up
               eos
             @env.ui.error(message)
-            exit 1
+            exit 2
```
bexelbie: The comment in this function may need to be updated.

I believe this is not an exit 2.  It either needs to be 3 (unreserved) or 1, imho.  Running this command without the machine running feels like an impermissible operation error.

lib/vagrant-service-manager/command.rb
```diff
@@ -41,16 +41,16 @@ def execute
                 # display information about all the providers inside ADB/CDK
                 self.print_all_provider_info
             else
-                self.print_help
+                self.print_help(1)
```
bexelbie, line 65: I believe help should exit with a 0.

see

```
ls;echo $?
ls --help;echo $?
```

lib/vagrant-service-manager/command.rb
```diff
@@ -41,16 +41,16 @@ def execute
...
             end
         when "help"
             self.print_help
```
bexelbie, line 71: See line 44

lib/vagrant-service-manager/command.rb
```diff
@@ -41,16 +41,16 @@ def execute
...
         else
-            self.print_help
+            self.print_help(1)
```
bexelbie, line 73: see line 44

lib/vagrant-service-manager/command.rb
```diff
@@ -109,8 +110,8 @@ def execute_openshift_info
           openshift_console_url = "#{openshift_url}/console"
           self.print_openshift_info(openshift_url, openshift_console_url)
         else
-          @env.ui.info("# OpenShift service is not running in the vagrant box.")
-          exit 1
+          @env.ui.error("# OpenShift service is not running in the vagrant box.")
+          exit 2
```
bexelbie: I vote for 126 here.

bexelbie:
I've added some comments inline with the script.

This PR cannot be merged without docs and a changelog update.

lib/vagrant-service-manager/command.rb
navidshaikh, line 65: if user has entered wrong command which we do not support in sub-commands, it should be non zero exit status.

```
ls --bex ; echo $?
```

lib/vagrant-service-manager/command.rb
navidshaikh: Good catch, will update the PR and description.

lib/vagrant-service-manager/command.rb
```diff
@@ -21,7 +21,7 @@ def exit_if_machine_not_running
   vagrant up
               eos
             @env.ui.error(message)
-            exit 1
+            exit 3
```
navidshaikh, line 24: @bexelbie : Updated the exit code and comment.

lib/vagrant-service-manager/command.rb
```diff
@@ -109,8 +110,8 @@ def execute_openshift_info
           openshift_console_url = "#{openshift_url}/console"
           self.print_openshift_info(openshift_url, openshift_console_url)
         else
-          @env.ui.info("# OpenShift service is not running in the vagrant box.")
-          exit 1
+          @env.ui.error("# OpenShift service is not running in the vagrant box.")
+          exit 126
```
navidshaikh, line 143: @bexelbie Updated the exit code and PR description.

lib/vagrant-service-manager/command.rb
bexelbie, line 65: Ok, I understand the exit code of 1 here.

Also, please stop making fun of my favorite ls option :P

lib/vagrant-service-manager/command.rb
bexelbie, line 71: For clarity here should we put self.print_help(0) ?

navidshaikh, line 71: Since user has given wrong command, it will exit with 1 and show the help.

navidshaikh:
Re-based and updated the PR with correct exit codes as per PR description.

navidshaikh:
ping @bexelbie @LalatenduMohanty @budhrg , please review.

bexelbie:
@navidshaikh this still needs docs and a changelog update

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
-## Unreleased
 - Enhanced the developer instructions for developing the plugin in README
```
bexelbie: Can you add @budhrg to this line as it got ommitted

README.md
```diff
@@ -32,6 +32,17 @@ The [Atomic Developer Bundle](https://github.com/projectatomic/adb-atomic-develo
 
 4. Begin using your host-based tools.
 
+## Exit codes
+
+Following table shows the exit codes from plugin and their meanings
```
bexelbie: The following table lists the plugin's exit codes and their meaning:

README.md
```diff
@@ -32,6 +32,17 @@ The [Atomic Developer Bundle](https://github.com/projectatomic/adb-atomic-develo
...
+
+Exit Code Number | Meaning
+---------------  |-------------------------------------------------------------------------
+1                | Catch all for general errors / Wrong sub-command or option given
```
bexelbie, line 42: Add:

0 | No Error

README.md
```diff
@@ -32,6 +32,17 @@ The [Atomic Developer Bundle](https://github.com/projectatomic/adb-atomic-develo
...
+3                | If vagrant box is not running and plugin is executed
```
bexelbie: The ADB is not running and must be before this command can succeed

CHANGELOG.md
navidshaikh: I will update it when sending full list of changelog

bexelbie:
`bundle exec vagrant service-manger env` returns an error message about openshift not running and exits with error code 126.  It shouldn't

print_all_provider_info assumes all providers are loaded and shouldn't.

README.md
navidshaikh: I'll vote for "Vagrant box" instead of "ADB"

navidshaikh:
> bundle exec vagrant service-manger env returns an error message about openshift not running and exits with error code 126. It shouldn't

Filed an issue for this https://github.com/projectatomic/vagrant-service-manager/issues/75

README.md
bexelbie: that is fine

bexelbie:
LGTM

Issue #72, opened 2016-02-24
Check if we can add a private networking for VirtualBox provider


navidshaikh:
Users have to configure the `private_networking` in Vagrantfile for VirtualBox providers.
Evaluate if plugin can configure the `private_networking` even if there is no mention of it in Vagrantfile and VirtualBox provider is used.

If we can get there, the user experience for ADB would be

```
vagrant init projectatomic/adb
vagrant up
```

LalatenduMohanty:
@navidshaikh this sounds awesome!. I will try to figure this out

LalatenduMohanty:
This has to be done in the provisioning step as adb-utils will use the interface to configure OpenShift.
CC @navidshaikh @praveenkumar @budhrg

LalatenduMohanty:
Or with the action hook , during the `vagrant up`

navidshaikh:
> This has to be done in the provisioning step as adb-utils will use the interface to configure OpenShift.

Yes, this is happening as part of action hook during `vagrant up` and before we start configuring services (docker, openshift).

bexelbie:
This is completed.  https://github.com/projectatomic/vagrant-service-manager/blob/master/lib/vagrant-service-manager/action/setup_network.rb#L13

PR #86, opened 2016-03-02
Enable private network for VirtualBox if not set

### LINKED ISSUE #16: Show warning if private networking is not configured for ADB
_From @navidshaikh on November 24, 2015 9:38_

For VirtualBox provider, private networking needs to be configured for ADB by adding following line in the `Vagrantfile`

```
```

if private networking is not configured, adbinfo output does not display the IP address. A proper warning message needs to be displayed giving tip about possible fix.

_Copied from original issue: projectatomic/vagrant-adbinfo#41_

brgnepal:
Fix #16 and #72.

Tested on `VirtualBox` and `Libvirt` provider locally. Working as expected.
For `VirtualBox`, we get network info as info highlighted between
 `*****************************************************`  below:

```
#### Private Networking for VirtualBox on "Vagrant Up" (line no 11)

budhram@dhcp35-83 ~/r/vagrant-service-manager (fix-72)> bundle exec vagrant up

Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'projectatomic/adb'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'projectatomic/adb' is up to date...
==> default: Setting the name of the VM: vagrant-service-manager_default_1456840362014_99026
==> default: Clearing any previously set network interfaces...
*********************************************************************************************************
==> default: Virtualbox requires an additional private network. Adding it...
*********************************************************************************************************
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
    default: Adapter 2: hostonly
==> default: Forwarding ports...
    default: 22 (guest) => 2222 (host) (adapter 1)
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
    default: Warning: Remote connection disconnect. Retrying...
    default: 
    default: Vagrant insecure key detected. Vagrant will automatically replace
    default: this with a newly generated keypair for better security.
    default: 
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if it's present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
    default: No guest additions were detected on the base box for this VM! Guest
    default: additions are required for forwarded ports, shared folders, host only
    default: networking, and more. If SSH fails on this machine, please install
    default: the guest additions and repackage the box to continue.
    default: 
    default: This is not an error message; everything may continue to work properly,
    default: in which case you may ignore this message.
==> default: Configuring and enabling network interfaces...
==> default: Rsyncing folder: /home/budhram/redhat/vagrant-service-manager/ => /vagrant
```

### DIFF
--- a/lib/vagrant-service-manager/action/setup_network.rb
+++ b/lib/vagrant-service-manager/action/setup_network.rb
@@ -1,0 +1,38 @@
+module Vagrant
+  module ServiceManager
+    module Action
+      class SetupNetwork
+
+        def initialize(app, env)
+          @app = app
+          @machine = env[:machine]
+          @ui = env[:ui]
+        end
+
+        def call(env)
+          add_private_network if virtualbox? && default_network_exists?
+          @app.call(env)
+        end
+
+        private
+
+        def virtualbox?
+          @machine.provider.instance_of?(VagrantPlugins::ProviderVirtualBox::Provider)
+        end
+
+        def default_network_exists?
+          @machine.config.vm.networks.length == 1
+        end
+
+        def add_private_network
+          @ui.info <<-MSG
+When using virtualbox, a non-NAT network interface is required.
+Adding a private network using DHCP
+          MSG
+          @machine.config.vm.network :private_network, type: :dhcp
+        end
+
+      end
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Fix links, typos, formatting in CONTRIBUTING.md @budhrg
+- Fix #16 and #72: Enable private networking for VirtualBox if not set @budhrg
 
 ## v0.0.3 Mar 01, 2016
 - Fix #74: vagrant-service-manager plugin version 0.0.3 release @navidshaikh
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -13,8 +13,6 @@ Vagrant.configure(2) do |config|
 
   config.vm.box = "projectatomic/adb"
 
-  config.vm.network "private_network", type: "dhcp"
-
   # This is the default setup
   # config.servicemanager.services = 'docker'
 
--- a/lib/vagrant-service-manager/plugin.rb
+++ b/lib/vagrant-service-manager/plugin.rb
@@ -1,5 +1,6 @@
-# Loads all services
+# Loads all services and actions
 Dir["#{File.dirname(__FILE__)}/services/*.rb"].each { |f| require_relative f }
+Dir["#{File.dirname(__FILE__)}/action/*.rb"].each { |f| require_relative f }
 
 module Vagrant
   module ServiceManager
@@ -18,7 +19,14 @@ module Vagrant
       end
 
       action_hook(:servicemanager, :machine_action_up) do |hook|
+        hook.before(VagrantPlugins::ProviderVirtualBox::Action::Network, setup_network)
         hook.append(Service::Docker)
+      end
+
+      def self.setup_network
+        Vagrant::Action::Builder.new.tap do |b|
+          b.use Action::SetupNetwork
+        end
       end
     end
   end


lib/vagrant-service-manager/action/setup_network.rb
```diff
@@ -0,0 +1,37 @@
...
+        end
+
+        private
+
+        def virtualbox?
+          @machine.provider.instance_of?(VagrantPlugins::ProviderVirtualBox::Provider)
+        end
+
+        def private_network_exists?
+          @machine.config.vm.networks.any? { |type, _| type == :private_network }
```
bexelbie: For virtualbox, I believe we should just detect if there are any networks added beyond the default.  The goal isn't to have a private network, but it is to have an interface that is not a NAT.

lib/vagrant-service-manager/action/setup_network.rb
```diff
@@ -0,0 +1,37 @@
...
+        end
+
+        def add_prerequisite_network
+          return unless virtualbox? && !private_network_exists?
+
+          @ui.info 'Virtualbox requires an additional private network. Adding it...'
```
bexelbie: Suggested:

When using virtualbox, a non-NAT network interface is required.
Adding a private network using DHCP

---

Can we get this to highlight inthe output?

navidshaikh:
apart from @bexelbie comments, this PR will need rebase.

lib/vagrant-service-manager/action/setup_network.rb
hferentschik: Is a DHCP interface really what we want? Should it not be the default IP we are using 10.1.2.2.
This should be paired with the ability to change the private network IP via a config option. With this in place we are again on step closer to getting fully out of the box experience for the user.

hferentschik:
On that note, I think the plugin should a) check whether a private network exist and b) add one (the default 10.1.2.2) if it doesn't. 

This should be paired with a config option `vagrant.servicemanager.ip` which can be used to configure the default private network IP. The default would be 10.1.2.2, but could be changed to get OpenShift et all running against a different IP.

With this in place, we can remove the requirement of adding manually a private network IP to the Vagrantfile and get a better out of the box experience (think `vagrant init cdk2 && vagrant up`).

lib/vagrant-service-manager/action/setup_network.rb
LalatenduMohanty: @hferentschik DHCP would be a welcome step. As users run multiple instance of the Vagrantbox (usually two) and Vagrant complains about it as the IP can not be given to two vagrantbox. So if service-manager can configure the private networking with dhcp then adb-utils scripts can pick the IP address and use it for deploying OpenShift.

lib/vagrant-service-manager/action/setup_network.rb
```diff
@@ -0,0 +1,37 @@
...
+          @machine.config.vm.network :private_network, type: :dhcp
```
LalatenduMohanty, line 32: Yw! so cool :-)

navidshaikh:
> This should be paired with a config option vagrant.servicemanager.ip which can be used to configure the default private network IP. The default would be 10.1.2.2, but could be changed to get OpenShift et all running against a different IP.

Why do we need to set default IP 10.1.2.2 ? If an IP can be allocated via private_network and displayed via plugin when request for openshift info.

navidshaikh:
> With this in place, we can remove the requirement of adding manually a private network IP to the Vagrantfile and get a better out of the box experience (think vagrant init cdk2 && vagrant up).

Indeed, its the goal!

lib/vagrant-service-manager/action/setup_network.rb
hferentschik: > As users run multiple instance of the Vagrantbox (usually two) 

Do they? why? What's the use case here?

> So if service-manager can configure the private networking with dhcp then adb-utils scripts can pick the IP address and use it for deploying OpenShift

I was not so much concerned about the provisioning step, but was more thinking about how the user knows the right IPs and URLs. If we have a "hard coded" default, users might have it easier to get started. On the other hand, with service-manager they can just run `vagrant service-manager env docker|openshift` and get the info they need. I guess that is how `docker-machine` works as well. 

Maybe we would still need a `vagrant service-manager ip` which returns the DHCP private network IP (independent of any specific service being enabled). Actually I quite like this idea. It's a step away from the default `10.1.2.2` approach we have been using. We might want to float this idea on a broader audience to get some feedback. WDYT?

hferentschik:
> Why do we need to set default IP 10.1.2.2 ? If an IP can be allocated via private_network and displayed via plugin when request for openshift info.

See my comment here - https://github.com/projectatomic/vagrant-service-manager/pull/86#discussion_r54716045
I am coming around the DHCP idea. But as said, it is a deviation of what we have done so far. Also, it should still work with a fixed IP, in the case a user can only use a certain IP range.

So, I guess it would be, use private network IP as explicitly configured by us user and if none exist use DHCP. The documentation should be around the DHCP approach then, explaining how to retrieve the IP once the VM is up and how to force a specific IP.

lib/vagrant-service-manager/action/setup_network.rb
LalatenduMohanty: Sure. The user should be able to choose dhcp or a static IP for the private interface. +1 to spending this idea to container-tools

navidshaikh:
> Also, it should still work with a fixed IP, in the case a user can only use a certain IP range.

That could be a valid use case.

> So, I guess it would be, use private network IP as explicitly configured by us user and if none exist use DHCP. The documentation should be around the DHCP approach then, explaining how to retrieve the IP once the VM is up and how to force a specific IP.

As part of this PR, I would prefer to have the private_networking set up if it does not exist, and having a separate PR to provision explicit configured specific IP and respective flow as mentioned.

hferentschik:
> As part of this PR, I would prefer to have the private_networking set up if it does not exist, and having a separate PR to provision explicit configured specific IP and respective flow as mentioned.

So with this PR merged what exactly happens for the case where:

a) I have explicitly set a private network (DHCP or fixed IP)
b) no private network config at all

bexelbie:
@hferentschik As I understand this change you'll see this:
1) no perceptible change. Your network settings are respected. Nothing is added. 
2) a dhcp private network is added.

hferentschik:
> As I understand this change you'll see this...

Ok, that was my understanding as well. So this is really from a service-manager point of view the behavior we want, right? The rest would be documentation changes to and potentially a `vagrant service-manager ip` to use the DHCP approach as a default.

bexelbie:
@hferentschik I'm not sure I completely follow. `vagrant service-manager ip` should always return the IP of the service end point regardless of how the network got started.

hferentschik:
Well, first of there is no `vagrant service-manager ip` yet right? This is an additional command I am proposing. One can use `vagrant service-manager env <docker|openshift|...>` to get service specific information which will in one way or another contain the used private network IP.

However, I would find it quite useful to know just this private IP independent of a service. That's what I am proposing with `vagrant service-manager ip`. It this not a must, but imo a nice to have. It is the equivalent to `docker-machine ip <name>`. 

@bexelbie does this make more sense now?

bexelbie:
@hferentschik The IP subcommand has already been discussed internally. I didn't realize we hadn't said anything publicly. It is planned :)

hferentschik:
> The IP subcommand has already been discussed internally. I didn't realize we hadn't said anything publicly. It is planned :)

nice

brgnepal:
@bexelbie , @hferentschik . Updated as:
1. If only default network(NAT), then add private dhcp network.
2. If user set any network in Vagrantfile, use that one.

navidshaikh:
LGTM

navidshaikh:
> This should be paired with a config option vagrant.servicemanager.ip which can be used to configure the default private network IP. The default would be 10.1.2.2, but could be changed to get OpenShift et all running against a different IP.

I think we should not introduce `vagrant.servicemanager.ip` config option, with this option we are introducing a new way to give private network IP to box, IMO `config.vm.network "private_network", ip: "10.1.2.2"` is better known way for vagrant users.

bexelbie:
@hferentschik @navidshaikh maybe I mis understood.  I saw `vagrant service-manager ip` as a way to get the ip of the box, not set it.  But yes, it belongs in a separate PR

hferentschik:
> I think we should not introduce vagrant.servicemanager.ip config option, with this option we are introducing a new way to give private network IP to box, IMO config.vm.network "private_network", ip: "10.1.2.2" is better known way for vagrant users. 

fair enough

hferentschik:
>  I saw vagrant service-manager ip as a way to get the ip of the box, not set it

It is, but we also discussed an actual option `vagrant.servicemanager.ip` to set the private IP. But as @navidshaikh is saying people are probably more familiar with `config.vm.network`.

> But yes, it belongs in a separate PR

+1

bexelbie:
LGTM

PR #88, opened 2016-03-02
Supports starting OpenShift service as part of config

### LINKED ISSUE #83: Refactor lib/vagrant-service-manager/services/docker.rb to move box detection up a level
In preparation for an openshift (and others) service provisioning, move the detection of the ADB/CDK boxes up a level so that the code is done before the service provisioners are called, not in the provisioners themselves.

Ideal workflow would allow a new service to be added solely by creating a new file in lib/vagrant-service-manager/services/ that doesn't need to know anything but how to start, stop, and provision the service in question.

@budhrg can you do this?

### LINKED ISSUE #87: Make OpenShift enabled per default
According to the docs, OpenShift needs to be enabled explicitly via `config.servicemanager.services = 'openshift'`. I was under the assumption that Openshift is the default and does not need to be enabled explicitly.

navidshaikh:
Fixes #87:
 Fixes #83:
- Adds support for enabling and starting OpenShift service
- For CDK, OpenShift service is enabled and started by default as part of vagrant up
- For ADB, OpenShift service needs to be configured in Vagrantfile as
  config.servicemanager.services = 'openshift'


navidshaikh:
I tested this PR against CDK and ADB.
For CDK it works fine as expected.
For ADB, I am getting following error

```
➜  vagrant-service-manager git:(fix-87) ✗ bundle exec vagrant up --provider virtualbox
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'projectatomic/adb'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'projectatomic/adb' is up to date...
==> default: Setting the name of the VM: vagrant-service-manager_default_1456924885101_37932
==> default: Fixed port collision for 22 => 2222. Now on port 2201.
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
    default: Adapter 2: hostonly
==> default: Forwarding ports...
    default: 22 (guest) => 2201 (host) (adapter 1)
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2201
    default: SSH username: vagrant
    default: SSH auth method: private key
    default: 
    default: Vagrant insecure key detected. Vagrant will automatically replace
    default: this with a newly generated keypair for better security.
    default: 
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if it's present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
    default: No guest additions were detected on the base box for this VM! Guest
    default: additions are required for forwarded ports, shared folders, host only
    default: networking, and more. If SSH fails on this machine, please install
    default: the guest additions and repackage the box to continue.
    default: 
    default: This is not an error message; everything may continue to work properly,
    default: in which case you may ignore this message.
==> default: Job for openshift.service failed because the control process exited with error code. See "systemctl status openshift.service" and "journalctl -xe" for details.
```

Tried `vagrant ssh` and starting the service manually, still failing

```
$ vagrant ssh
Last login: Wed Mar  2 08:33:10 2016 from 10.0.2.2
[vagrant@centos7-adb ~]$ sudo systemctl start openshift
Job for openshift.service failed because the control process exited with error code. See "systemctl status openshift.service" and "journalctl -xe" for details.

```

lib/vagrant-service-manager/services/openshift.rb
```diff
@@ -0,0 +1,37 @@
...
+            elsif @services.include? "openshift"
+              # Its ADB and start openshift service if it is configured in Vagrantfile
+              self.start_enable_openshift
+            end
+          end
+          @app.call(env)
+        end
+
+        def start_enable_openshift
+          command = "sudo systemctl enable openshift && sudo systemctl restart openshift"
```
bexelbie: This needs to use sccli not systemctl.

lib/vagrant-service-manager/plugin.rb
```diff
@@ -19,6 +19,7 @@ class Plugin < Vagrant.plugin('2')
 
       action_hook(:servicemanager, :machine_action_up) do |hook|
         hook.append(Service::Docker)
+        hook.append(Service::OpenShift)
```
bexelbie: This needs to be if-def'ed off so that OpenShift is default for RH CDK only.  We don't need hte docker append if we do OpenShift, afaik.

lib/vagrant-service-manager/services/openshift.rb
bexelbie: The command should specifically either setup an environment for the shell to be used or call the command with explicit variables.  No variable should be passed unless it is provided by the user.  See the mods to README.md for notes.

`sudo DOCKER_REGISTRY=#{DOCKER_REGISTRY} IMAGE_TAG=#{IMAGE_TAG} IMAGE_NAME=#{IMAGE_NAME} /usr/bin/sccli openshift`

The variables are:
- DOCKER_REGISTRY
- IMAGE_NAME
- IMAGE_TAG

navidshaikh:
rebased and updated the PR as per suggestions.

lib/vagrant-service-manager/service.rb
```diff
@@ -0,0 +1,34 @@
...
+        @machine = env[:machine]
+        @ui = env[:ui]
+        @services = @machine.config.servicemanager.services.split(',').map(&:chomp)
+        @docker_hook = Docker.new(@machine, @ui)
+        @openshift_hook = OpenShift.new(@machine, @ui)
+      end
+
+      def call(env)
+        if SUPPORTED_BOXES.include? @machine.guest.capability(:flavor)
+          @docker_hook.execute
```
bexelbie, line 22: It seems that there should be a check here.  I should be allowed to say I don't want Docker (or later OpenShift).

lib/vagrant-service-manager/services/openshift.rb
```diff
@@ -0,0 +1,24 @@
+module Vagrant
+  module ServiceManager
```
bexelbie: If this service requires docker, shouldn't we somehow have it call docker in case the user didn't already specify it?  It should also detect if docker was listed and not re-call it.

bexelbie:
Let's squash before merge after other issues are resolved.

lib/vagrant-service-manager/service.rb
brgnepal, line 22: @bexelbie : Isn't docker is to be run by default? Like any time when machine up.

brgnepal:
@navidshaikh, @bexelbie, @LalatenduMohanty  :  This PR has `flavor` thing that is updated on the new PR https://github.com/projectatomic/vagrant-service-manager/pull/92.

Let's first merge PR#92, then update this PR and merge.

lib/vagrant-service-manager/services/openshift.rb
navidshaikh: Irrespective of whether docker is being listed or not, or a service (openshift) depends on docker, since we are enabling docker by default, its implicit. Why do we need to link it in dependency chain ?

hferentschik: I even wonder why we talk about Docker as a service. OpenShift, Kubernetes, Atomic etc are for me the services. Docker is needed for all of them afaik. A ADB/CDK w/o Docker does not make any sense whatsoever. My standpoint is that Docker configuration is always implicit and just happens and depending whether additional services are enabled there might be additional  Docker options set, but overall Docker is just a process running on the VM. I think taking this POV makes things less ambiguous.

lib/vagrant-service-manager/service.rb
```diff
@@ -0,0 +1,34 @@
...
+        @machine = env[:machine]
+        @ui = env[:ui]
+        @services = @machine.config.servicemanager.services.split(',').map(&:chomp)
+        @docker_hook = Docker.new(@machine, @ui)
+        @openshift_hook = OpenShift.new(@machine, @ui)
+      end
+
+      def call(env)
+        if SUPPORTED_BOXES.include? @machine.guest.capability(:os_variant)
+          @app.call(env)
```
brgnepal: This should be the first line in this method.

If our box is not supported then, it will break the middleware chain.

README.md
```diff
@@ -36,6 +36,17 @@ producing complex, multi-container applications.
    * `docker` is default service and does not require above configuration.
    * Enable multiple services as comma separated list. Eg: 'docker, openshift'
 
+4. Enable any specific options for the services you have selected:
+
+    * OpenShift
+
+        Specific versions can be specified using the following variables:
+
+        `config.servicemanager.openshift.docker_registry = "docker.io"` - What registry should be pulled from
```
hferentschik: Where is the code which takes care of these properties? Aka makes sure the right things happen when these properties are set?

navidshaikh: @hferentschik : The feature is being implemented in chunks with testing. ATM there is no code to take care of it, but README is our TODO to look at and implement. :-)

hferentschik: @navidshaikh Not quite sure whether the README makes a good TODO. If I visit the repo and read the readme, I would expect the things mentioned there to be implemented. Maybe not releases, but I'd install from source I'd expect it to work. IMO either use github issues or wiki page or utilize the TODO file which already exists.

brgnepal:
@bexelbie , update code.
Full command looks like 

```
sudo DOCKER_REGISTRY='docker.io' IMAGE_NAME='openshift/origin' IMAGE_TAG='v1.1.1' sccli openshift
```

bexelbie:
This seems to be working, however once everything is started I can't log into OpenShift.  I can't decide what the source of the TLS error is.  I suspect this isn't in this plugin, but I don't see it.  @budhrg @praveenkumar @navidshaikh 

Also, this may not be starting enough containers on CDK.

brgnepal:
@bexelbie , I also hit by the same `TLS Error`.

```
.....
 ==> default: Error while pulling image: Get https://registry.access.redhat.com/v1/repositories/openshift3/ose/images: net/http: TLS handshake timeout
```

praveenkumar:
@budhrg @bexelbie which command options are you using to get this issue? I tried in cdk box to pull a ose image tag and it's working without any `TLS Error`.

brgnepal:
@praveenkumar :  Following is the comand.

sudo DOCKER_REGISTRY='docker.io' IMAGE_NAME='openshift/origin' IMAGE_TAG='v1.1.1' sccli openshift

Is single quote fine here? 
I see there is no single quote in [this command](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/components/centos/centos-openshift-setup/Vagrantfile#L65)

bexelbie:
@praveenkumar This sequence:

vagrant up
Vagrant service-manager env OpenShift
Oc login

navidshaikh:
I tested the PR with
- explicitly configuring the openshift with default options
- using CDK (where openshift is default) and no explicit configuration mentioned with default options
- using CDK and no explicit configuration mentioned with custom options for  docker registry, image name and image tag.

Works fine! :-)

praveenkumar:
Can you folks squash all comments in single one?

navidshaikh:
> Can you folks squash all comments in single one?

Different commits in this PR are solving sub-features of the entire feature. I'll vote for having the individual commits in this PR.

navidshaikh:
> @navidshaikh , @bexelbie : Probably written as:

I am following the Vagrant style of specifying config options in comment. For eg if you init a box using `vagrant init projectatomic/adb` and you check the Vagrantfile you can see

```
  # Create a private network, which allows host-only access to the machine
  # using a specific IP.
  # config.vm.network "private_network", ip: "192.168.33.10"
```

navidshaikh:
LGTM!

brgnepal:
LGTM!

lib/vagrant-service-manager/service.rb
```diff
@@ -0,0 +1,35 @@
...
+        @openshift_hook = OpenShift.new(@machine, @ui)
+      end
+
+      def call(env)
+        @app.call(env)
+
+        if SUPPORTED_BOXES.include? @machine.guest.capability(:os_variant)
+          @docker_hook.execute
+
+          if @machine.guest.capability(:os_variant) == "cdk"
```
bexelbie: We should only execute here if there is no services line.  We should not force openshift on CDK, only run it as default.

lib/vagrant-service-manager/service.rb
```diff
@@ -0,0 +1,35 @@
...
+        @ui = env[:ui]
+        @services = @machine.config.servicemanager.services.split(',').map(&:strip)
+        @docker_hook = Docker.new(@machine, @ui)
+        @openshift_hook = OpenShift.new(@machine, @ui)
+      end
+
+      def call(env)
+        @app.call(env)
+
+        if SUPPORTED_BOXES.include? @machine.guest.capability(:os_variant)
```
bexelbie, line 21: Can we check specifically for CDK and ADB variants here?

praveenkumar:
LGTM

lib/vagrant-service-manager/service.rb
praveenkumar: @bexelbie I think that was the plan to include openshift as default service by default in CDK but if a user defined other service explicitly then that service should take preference.

brgnepal: @bexelbie Does it mean that the line in Vagrantfile drvies the execution of openshift? 

I assume this:
1. If no service line -->  Execute only docker
2. If service line with "openshift" --->  Execute docker and openshift
3. If service line with "docker, openshift" ---> Execute docker and openshift

Similar for kubernetes.
So here variant will not decide which service to start.
Right?

lib/vagrant-service-manager/service.rb
navidshaikh, line 21: @bexelbie : We are doing the same thing here. In line no:6 of this file we are defining this

```
SUPPORTED_BOXES = ['adb', 'cdk']
```

and the guest is checked if it is subset of `SUPPORTED_BOXES`.

Did I answer your question or you wanted to ask something else?

brgnepal, line 21: @bexelbie : This line will only prevent other boxes not to trigger this plugin.
I would like to know your use case here please.

lib/vagrant-service-manager/service.rb
navidshaikh: > We should only execute here if there is no services line. We should not force openshift on CDK, only run it as default.

The only two supported services we can configure for user is `docker` and `openshift`. `docker` service is configured by default irrespective of the box in action, which leaves us to `openshift`. And since `openshift` to be enabled by default for CDK we are enabling it by default irrespective of the explicit mention in Vagrantfile (as for docker).

Later, when we'll have support for more services (kubernetes, mesos) the logic should be expanded to handle the collision and check for explicit mention of the services.

Given the scope of feature we have for this release, I think this solves the scoped problem in hand for the release.

bexelbie: I understand it to be like this

If no service line and ADB - execute docker
if no service line and CDK - execute openshift
if service line execute what is on service line
otherwise do nothing

lib/vagrant-service-manager/service.rb
bexelbie, line 21: Misread the line :)

lib/vagrant-service-manager/service.rb
bexelbie: @navidshaikh I should be able to start a CDK box without openshift if I so desire

brgnepal: > otherwise do nothing

@bexelbie , @navidshaikh : this looks like contradicting to what we were initially thought before implemented feature for docker(docker restart), now openshift. Is it something new?

navidshaikh: I understand it to be like this

> If no service line and ADB - execute docker
> if no service line and CDK - execute openshift
> if service line execute what is on service line
> otherwise do nothing

The same behavior is there. Did you see any deflection ?

bexelbie: Yes.  Service line should control what happens.  Defaults apply in the absence of a choice in the service line.

navidshaikh:
> What do you think about this in one line as?

@budhrg : in response to your comment about logic
I think we already have a check for `SUPPORTED_BOXES` and we do not need to add an extra check for `@services.include?("openshift")` since that is covered in next `if` block. Basically just reducing the number of checks and cycles of comparison.

bexelbie:
LGTM

PR #99, opened 2016-03-09
Fix #80 #6: Check for correct TLS certs pair

### LINKED ISSUE #6: plugin does not re-generate the certs in 'secrets path' docker directory
_From @surajssd on January 27, 2016 7:30_

When manually deleting certificates on host in _.docker_ directory, and re-running `vagrant adbinfo` does not re-generate the certificates.

Steps to re-create:

``` bash
$ ls .vagrant/machines/default/libvirt/.docker/
ca.pem  cert.pem  key.pem
$ rm -rf .vagrant/machines/default/libvirt/.docker/*.*
$ ls .vagrant/machines/default/libvirt/.docker/
$ vagrant adbinfo
export DOCKER_HOST=tcp://192.168.121.36:2376
export DOCKER_CERT_PATH=/home/hummer/atomic/workspace/.vagrant/machines/default/libvirt/.docker
export DOCKER_TLS_VERIFY=1
export DOCKER_MACHINE_NAME=5aae67f
# eval "$(vagrant adbinfo)"


$ ls .vagrant/machines/default/libvirt/.docker/
$ 
```

Note: Rather than deleting the files in the directory, if you delete the directory itself, certificates are generated.

_Copied from original issue: projectatomic/vagrant-adbinfo#78_

### LINKED ISSUE #80: Check for correct TLS certs pair
User story:
I am an ADB user, I `vagrant up` my box and using `vagrant-service-manager` plugin I copied the client side certs in my host directory. I `vagrant destroy` after completing my work (the copied over certs stays un-touched). The next day I `vagrant up` again (which will re-generate the certs inside the box), I try to copy the certs using `vagrant-service-manager`, however the certs will not be copied over since yesterday's certs are already there, and plugin will not copy, but just show the configuration info.

Problem:
Now there are incorrect client side certs on host. 

Solution:
plugin should ensure that the client side certs are correct and are part of the pair from the server side certs.

How does this work today:
`rm -rf .vagrant/machines/default/virtualbox/docker` manually before copying the certs using plugin to ensure I have correct client side certs.

brgnepal:
Fixes #80 
Fixes #6 
- Checks validity of certs if 'docker' folder exists in host

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -1,9 +1,12 @@
 require_relative 'os'
+require 'digest'
 
 module Vagrant
   module ServiceManager
     class Command < Vagrant.plugin(2, :command)
-      OS_RELEASE_FILE = "/etc/os-release"
+      OS_RELEASE_FILE = '/etc/os-release'
+      DOCKER_PATH = '/home/vagrant/.docker'
+
       def self.synopsis
         'provides the IP address:port and tls certificate file location for a docker daemon'
       end
@@ -155,34 +158,42 @@ Verb:
         end
       end
 
+      def sha_id(file_data)
+        Digest::SHA256.hexdigest file_data
+      end
+
+      def certs_present_and_valid?(path, machine)
+        return false if Dir["#{path}/*"].empty?
+
+        # check validity of certs
+        Dir[path + "/*"].each do |f|
+          guest_file_path = "#{DOCKER_PATH}/#{File.basename(f)}"
+          guest_sha = machine.guest.capability(:sha_id, guest_file_path)
+          return false if sha_id(File.read(f)) != guest_sha
+        end
+
+        true
+      end
+
       def execute_docker_info
         # this execute the operations needed to print the docker env info
         with_target_vms(nil, {:single_target=>true}) do |machine|
           # Path to the private_key and where we will store the TLS Certificates
           secrets_path = File.expand_path("docker", machine.data_dir)
 
-          hIP = machine.ssh_info[:host]
-          hport = machine.ssh_info[:port]
-          husername = machine.ssh_info[:username]
-
-          # Find the guest IP
-          guest_ip = self.find_machine_ip
-
           # Hard Code the Docker port because it is fixed on the VM
           # This also makes it easier for the plugin to be cross-provider
           port = 2376
 
-          # First, get the TLS Certificates, if needed
-          if !File.directory?(secrets_path) then
+          # Get the TLS certificates if needed
+          unless certs_present_and_valid?(secrets_path, machine)
+            Dir.mkdir(secrets_path) unless Dir.exists?(secrets_path)
 
-            # Create the directory
-            Dir.mkdir(secrets_path)
-
-            # copy the required client side certs from inside the box to host machine
+            # Copy the required client side certs from inside the box to host machine
             @env.ui.info("# Copying TLS certificates to #{secrets_path}")
-            machine.communicate.download("/home/vagrant/.docker/ca.pem", "#{secrets_path}")
-            machine.communicate.download("/home/vagrant/.docker/cert.pem", "#{secrets_path}")
-            machine.communicate.download("/home/vagrant/.docker/key.pem", "#{secrets_path}")
+            machine.communicate.download("#{DOCKER_PATH}/ca.pem", "#{secrets_path}")
+            machine.communicate.download("#{DOCKER_PATH}/cert.pem", "#{secrets_path}")
+            machine.communicate.download("#{DOCKER_PATH}/key.pem", "#{secrets_path}")
           end
 
           api_version = ""
@@ -192,7 +203,7 @@ Verb:
           end
 
           # display the information, irrespective of the copy operation
-          self.print_docker_env_info(guest_ip, port, secrets_path, machine.index_uuid, api_version)
+          self.print_docker_env_info(find_machine_ip, port, secrets_path, machine.index_uuid, api_version)
         end
       end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix #80: Check for correct TLS certs pair @budhrg
 - Fix #113: Adds DOCKER_API_VERSION in env docker output @navidshaikh
 
 ## v0.0.4 Mar 14, 2016
--- a/plugins/guests/redhat/cap/sha_id.rb
+++ b/plugins/guests/redhat/cap/sha_id.rb
@@ -1,0 +1,18 @@
+module VagrantPlugins
+  module GuestRedHat
+    module Cap
+      class ShaID
+        def self.sha_id(machine, path)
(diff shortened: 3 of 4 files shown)


lib/vagrant-service-manager/command.rb
```diff
@@ -172,6 +175,31 @@ def find_machine_ip
...
+          name = File.basename(f)
+
+          guest_file_path = "#{DOCKER_PATH}/#{name}"
+          machine.communicate.execute("cat #{guest_file_path}") do |type, data|
+            if type == :stderr
+              @ui.error(data)
+              exit 126
+            end
+
+            return false if sha_id(File.read(f)) != sha_id(data)
```
bexelbie: Should we do the sha of the docker cert on the VM?

brgnepal: @bexelbie : Didn't get you here. Could u elaborate more?

bexelbie: Right now you are downloading the cert from the VM and calculating the sha on in ruby.  Would it be more efficient to run the sha on the VM and just get that info?

navidshaikh: > Right now you are downloading the cert from the VM and calculating the sha on in ruby. Would it be more efficient to run the sha on the VM and just get that info?

+1

brgnepal: @bexelbie @navidshaikh Ok will more on its efficient way of doing it.

brgnepal:
@bexelbie , @navidshaikh : Updated vagrant way to get sha id for file.

navidshaikh:
I tested this PR, brought up a box , copied certs to host, malfunctioned the host machine (client side) certs and ran `vagrant service-manager env docker` again, even though host machine had the certs,  plugin checked for their shas and copied the certs again.

LGTM!

navidshaikh:
This also fixes #6

LalatenduMohanty:
@budhrg Please rebase the PR

lib/vagrant-service-manager/command.rb
```diff
@@ -155,6 +158,23 @@ def find_machine_ip
...
+      end
+
+      def certs_present_and_valid?(path, machine)
+        return false unless File.directory?(path)
+
+        # check validity of certs
+        Dir[path + "/*"].each do |f|
+          guest_file_path = "#{DOCKER_PATH}/#{File.basename(f)}"
+          guest_sha = machine.guest.capability(:sha_id, guest_file_path)
+          return false if sha_id(File.read(f)) != guest_sha
```
LalatenduMohanty: We should be comparing sha checksum between files in #{secrets_path} and #{DOCKER_PATH} here right ?

LalatenduMohanty:
@navidshaikh @budhrg have you guys tested this PR for #6? As it does not fix #6 for me.
Steps to reproduce for #6 

```
[lmohanty@LalatenduM-laptop adb]$ vagrant service-manager env docker
# Copying TLS certificates to /home/lmohanty/vagrantboxes/adb/.vagrant/machines/default/libvirt/docker
export DOCKER_HOST=tcp://192.168.121.125:2376
export DOCKER_CERT_PATH=/home/lmohanty/vagrantboxes/adb/.vagrant/machines/default/libvirt/docker
export DOCKER_TLS_VERIFY=1
export DOCKER_MACHINE_NAME=6210b46


[lmohanty@LalatenduM-laptop adb]$ ls .vagrant/machines/default/libvirt/docker/
ca.pem  cert.pem  key.pem
[lmohanty@LalatenduM-laptop adb]$ rm -rf .vagrant/machines/default/libvirt/docker/*
[lmohanty@LalatenduM-laptop adb]$ ls .vagrant/machines/default/libvirt/docker/
[lmohanty@LalatenduM-laptop adb]$ 
[lmohanty@LalatenduM-laptop adb]$ vagrant service-manager env docker
export DOCKER_HOST=tcp://192.168.121.125:2376
export DOCKER_CERT_PATH=/home/lmohanty/vagrantboxes/adb/.vagrant/machines/default/libvirt/docker
export DOCKER_TLS_VERIFY=1
export DOCKER_MACHINE_NAME=6210b46


[lmohanty@LalatenduM-laptop adb]$ ls .vagrant/machines/default/libvirt/docker/
[lmohanty@LalatenduM-laptop adb]$
```

lib/vagrant-service-manager/command.rb
```diff
@@ -155,6 +158,23 @@ def find_machine_ip
...
+        end
+
+        true
+      end
+
```
LalatenduMohanty, line 177: Because of some reason this method is not returning false while testing #6 .
My initial guess was may be `secrets_path = File.expand_path("docker", machine.data_dir)` does not have files (as these are deleted while I am testing #6) may be the logic does not work. But modifying code in that line did not give me positive result.

lib/vagrant-service-manager/command.rb
```diff
@@ -155,6 +158,23 @@ def find_machine_ip
...
+        return false unless File.directory?(path)
+
+        # check validity of certs
+        Dir[path + "/*"].each do |f|
+          guest_file_path = "#{DOCKER_PATH}/#{File.basename(f)}"
+          guest_sha = machine.guest.capability(:sha_id, guest_file_path)
+          return false if sha_id(File.read(f)) != guest_sha
+        end
+
+        true
```
LalatenduMohanty, line 175: What is the role of `true` here.

brgnepal, line 175: line no 169 to 173 are steps for validation. The first occurrence of invalidation out of 3 certs file will break the `each` loop and return false. If validation pass then `true` is returned from method.

LalatenduMohanty, line 175: @budhrg I think we should use `return true` instead of just `true` for better readability.

brgnepal, line 175: IMO without `return` is the better readability for the Ruby programmers at this line. `return` is there in ruby for scenario like line no 172.

lib/vagrant-service-manager/command.rb
brgnepal, line 177: I got the issue here. Will update the PR.

lib/vagrant-service-manager/command.rb
LalatenduMohanty, line 175: @budhrg  True, In Ruby it is advisable to not use `return` if not necessary. However IMO  if in a method if we have used explicit return already then we should continue the same style for consistency and readability.

brgnepal, line 175: I think it is not necessary as it is a general practice in Ruby to have such. 
@bexelbie @navidshaikh Your thoughts please.

lib/vagrant-service-manager/command.rb
brgnepal, line 177: @LalatenduMohanty Updated the PR now. It includes fix for #6 .

lib/vagrant-service-manager/command.rb
navidshaikh, line 175: The code LGTM, I'd vote for not having explicit `return`.

LalatenduMohanty:
LGTM

PR #100, opened 2016-03-11
enable to print export instead of setx in cygwin

### LINKED ISSUE #10: Print unix format of commands under cygwin
_From @optak on January 4, 2016 14:3_

adbinfo plugins print different hints for windows and non-windows system, but If working on windows but in cygwin, windows syntax is not much useful (setx variable value). At least I don't know how to use that variables in bash.
I believe it would be better to print "setx" examples only if running from cmd.exe (not bash in cygwin).

_Copied from original issue: projectatomic/vagrant-adbinfo#63_

optak:
Feature to enable bash style info in cygwin under windows.
Fixes #10 
If VAGRANT_DETECTED_OS=cygwin than print export instead of setx, but use windows-style path for cert (need to further usage).
Tested only on windows.

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -212,7 +212,7 @@ Verb:
 
         if !OS.windows? || OS.windows_cygwin? then
           if OS.windows? then
-            secrets_path = secrets_path.split('/').join('\\') + '\\'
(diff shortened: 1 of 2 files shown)


bexelbie:
LGTM - but untested @navidshaikh can you test this?

bexelbie:
Replaced by #145

PR #111, opened 2016-03-15
Fix-72: Add a private networking for VirtualBox

### LINKED ISSUE #16: Show warning if private networking is not configured for ADB

brgnepal:
Fix for #72 and #16 .

Explicitly checking for `private_network`. If not found, then set it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix-72: Add a private networking for VirtualBox @budhrg
 - Fix #113: Adds DOCKER_API_VERSION in env docker output @navidshaikh
 
 ## v0.0.4 Mar 14, 2016
--- a/lib/vagrant-service-manager/action/setup_network.rb
+++ b/lib/vagrant-service-manager/action/setup_network.rb
@@ -10,7 +10,7 @@ module Vagrant
         end
 
         def call(env)
-          add_private_network if virtualbox? && default_network_exists?
+          add_private_network if virtualbox? && !private_network_exists?
           @app.call(env)
         end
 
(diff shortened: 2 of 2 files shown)


navidshaikh:
I tested the PR, works fine as expected. I would like to test a scenario where user has configured a non-NAT / private network interface in Vagrantfile and we recognize it and do not add a network.

navidshaikh:
LGTM. @bexelbie ping do you have more scenarios in mind to test this PR against ?

LalatenduMohanty:
LGTM. 

> test a scenario where user has configured a non-NAT / private network interface in Vagrantfile and we recognize it and do not add a network.

+1

LalatenduMohanty:
One thing I am not sure is how we are handling private networking for libvirt? (Did not get a chance to look in to the code)

brgnepal:
@LalatenduMohanty : I think private network is not required for libvirt. Check https://github.com/projectatomic/vagrant-service-manager/issues/16. If we don't configure it in virtualbox, then we don't get IP in `DOCKER_HOST` env variable.

bexelbie:
@LalatenduMohanty technically we do not need to add a second network for libvirt because the first IP it gets automatically is routable.  However, adding one is not a problem and preserves the model that vagrant uses with other providers where the first IP is a "service" IP and the second+ are for services added by the user.

Not having the line on libvirt should still give an IP in hte env output.

bexelbie:
@navidshaikh @budhrg iirc we decided that this PR was unneeded and should be closed.  Please let me know if my recollection is wrong or close the PR.

brgnepal:
@bexelbie @navidshaikh ACK.

PR #114, opened 2016-03-16
Adds SPEC file for version 0.0.4 of the plugin

navidshaikh:
copr build https://copr.fedorainfracloud.org/coprs/nshaikh/vagrant-service-manager/build/168661/

### DIFF
--- a/vagrant-service-manager.spec
+++ b/vagrant-service-manager.spec
@@ -2,7 +2,7 @@
 %global vagrant_plugin_name vagrant-service-manager
 
 Name: %{vagrant_plugin_name}
-Version: 0.0.3
+Version: 0.0.4
 Release: 1%{?dist}
 Summary: To provide the user a CLI to configure the ADB/CDK for different use cases and to provide glue between ADB/CDK and the user's developer environment.
 Group: Development/Languages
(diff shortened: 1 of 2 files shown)


praveenkumar:
LGTM.

LalatenduMohanty:
@navidshaikh can you please rebase the PR?

navidshaikh:
@LalatenduMohanty : Rebased.

vagrant-service-manager.spec
```diff
@@ -69,6 +69,7 @@ popd
 %{vagrant_plugin_libdir}
 %exclude %{vagrant_plugin_cache}
 %{vagrant_plugin_spec}
+%{vagrant_plugin_instdir}/plugins/guests/
```
LalatenduMohanty, line 72: I am not sure if we have files in %{vagrant_plugin_dir}/\* related to plugins/guests . The question is do we really care about ruby code with respect to plugins/guests in the RPM. IMO if we have some files in %{vagrant_plugin_dir}/\* with respect to plugins/guests, we should remove them from the RPM.

praveenkumar, line 72: As per https://fedoraproject.org/wiki/PackagingDrafts/Vagrant#Macros %{vagrant_plugin_instdir} should contain actual content of the plugin like Gemfile, Rakefile, tests ..etc. Plugin gem should be part of %{vagrant_plugin_dir} and code content should be part of %{vagrant_plugin_libdir}

praveenkumar:
LGTM

Issue #98, opened 2016-03-08
vagrant service-manager should be runnable from any directory


hferentschik:
Given that the CDK provides a general purpose dev environment for potentially multiple projects on the host system, the user should be able to run service-manager commands in any directory. Atm one can only run the commands in the directory containing the _Vagrantfile_. In this case no additional parameter is needed. In other directories the user will need to specify the box id as determines by `vagrant global-status`, eg `vagrant service-manager env docker 32869cf`.

Some of the built-in Vagrant commands already allow for passing of the id. Also when service-manager is called outside of the directory containing the _Vagrantfile_ we currently get:

```
A Vagrant environment or target machine is required to run this
command. Run `vagrant init` to create a new Vagrant environment. Or,
get an ID of a target machine from `vagrant global-status` to run
this command on. A final option is to change to a directory with a
Vagrantfile and to try again. 
```

navidshaikh:
+1

navidshaikh:
> In other directories the user will need to specify the box id as determines by vagrant global-status, eg vagrant service-manager env docker 32869cf.

About providing the machine-id, I think specifying the machine-id as last parameter to command might be confusing, for options like `vagrant service-manager env docker --force-copy` (for forced copy of the certs from box to host, TBD).
I'll vote for having an option to each command as `--machine-id=32869cf`, if not given look for current working directory.
@bexelbie thoughts?

brgnepal:
+1 for `--machine-id` option as user will have clear understanding of it for "what to do with option?". 
Moreover, we can also document it well.

brgnepal:
Based on my discussion with @hferentschik , we have come up with following ideas:
- When ADB vm starts, the service-manager hook get executes which writes some metadata into `~/.vagrant/data/service-manager` folder. Metadata will be name of the box (eg adb/cdk) and the vagrant machine id. Vagrant allows to run commands outside the Vagrantfile directory but one need to specify the machine id.
- Optionally, we could have similar list as `vagrant global-status --prune` provides as
  `$ vagrant service-manager list machines`. It will list all ADB/CDK machines availables.
- On getting machine id, user will access plugin as:

```
$ vagrant service-manager env docker [--name <vm-name>]
$ vagrant service-manager box version [--name <vm-name>]
```

_name_ can be provided by user in config. service-manager maps from _name_ to Vagrant machine id. Without a name some sort of default is assumed, eg _adb_/_cdk_.

bexelbie:
[...]

PR #123, opened 2016-03-21
Fix #121: Removes DOCKER_MACHINE_NAME from `env docker` command output

### LINKED ISSUE #121: Remove DOCKER_MACHINE_NAME from service-manager env docker call
There seems to be no reason to have this variable in the generated output. At least not for the Host to VM docker communication. See also https://github.com/projectatomic/vagrant-service-manager/issues/21#issuecomment-199194003

navidshaikh:
Fixes #121

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix #121: Removes DOCKER_MACHINE_NAME from `env docker` command output @navidshaikh
 - Fix #65: Adds --script-readable option for `env openshift` command @navidshaikh
 - Fix #80: Check for correct TLS certs pair @budhrg
 - Fix #113: Adds DOCKER_API_VERSION in env docker output @navidshaikh
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -219,11 +219,11 @@ OPENSHIFT_WEB_CONSOLE=#{openshift_console_url}
           end
 
           # display the information, irrespective of the copy operation
-          self.print_docker_env_info(find_machine_ip, port, secrets_path, machine.index_uuid, api_version)
+          self.print_docker_env_info(guest_ip, port, secrets_path, api_version)
         end
       end
 
-      def print_docker_env_info(guest_ip, port, secrets_path, machine_uuid, api_version)
+      def print_docker_env_info(guest_ip, port, secrets_path, api_version)
         # Print configuration information for accesing the docker daemon
 
         if !OS.windows? then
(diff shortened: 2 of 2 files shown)


hferentschik:
+1 from my side. As said technically, this variable is not needed from a Docker communication point of view. In fact when it comes to docker-machine it was also added quite late. I found this discussion - https://github.com/docker/machine/issues/925. Exporting the name, helps to distinguish different Docker connection in separate consoles, eg via using the name in the prompt. The difference here is, that in the case of docker-machine we are actually dealing with human assigned names, which can indeed help. In our case we are just writing out the Vagrant machine id which is not so helpful.

In the long run it makes sense to be able to control the CDK(s) from another directory than the one the _Vagrantfile_ is located. At that stage one would need a name/id. But right now we don't have even this functionality and I think we need to think through first on how such a feature would look like. For example, could we also assign/use human readable name.

Bottom line, as long as we only have one CDK and we can only control it from within the directory the _Vagrantfile_ resides, we are imo better off not having this variable at all and introduce it when we have a need for it.

navidshaikh:
> In our case we are just writing out the Vagrant machine id which is not so helpful.

Once we have #98 implemented, having the machine id in the output will help since the machine id displayed is same as one sees in `vagrant global-status` output. 

> Bottom line, as long as we only have one CDK and we can only control it from within the directory the Vagrantfile resides, we are imo better off not having this variable at all and introduce it when we have a need for it.

+1

@hferentschik 
Also I would like to confirm other tooling parsing the output of `env docker` should not break if we have changed bits in the output.

navidshaikh:
@robstryker ping could you  also check? here is the  updated output

```
$vagrant service-manager env docker
# Copying TLS certificates to /home/nshaikh/work/src/vagrant-service-manager/.vagrant/machines/default/virtualbox/docker
export DOCKER_TLS_VERIFY=1

```

hferentschik:
> Once we have #98 implemented, having the machine id in the output will help since the machine id displayed is same as one sees in vagrant global-status output. 

Sure, but what I am saying here is that if possible we should be able to use a string id provided by the user instead of a bunch of numbers. That's what docker-machine does as well. Having a VM name of _local_cdk_ is helpful and easy. Having to use 953134, not so much. We need to investigate how we can use a user specified name.

navidshaikh:
> but what I am saying here is that if possible we should be able to use a string id provided by the user instead of a bunch of numbers. That's what docker-machine does as well. 

Understood!

> Having a VM name of local_cdk is helpful and easy. Having to use 953134, not so much. We need to investigate how we can use a user specified name.

+1 here is the tracker for same https://github.com/projectatomic/vagrant-service-manager/issues/124

robstryker:
JBT has not made use of DOCKER_MACHINE_NAME in any form, to the best of my knowledge.

brgnepal:
LGTM

navidshaikh:
@bexelbie : This is ready to be merged.

praveenkumar:
LGTM

Issue #122, opened 2016-03-21
Docker certificates should be copied to host at the time when they are generated


hferentschik:
and not at the first time `env docker` is called.

navidshaikh:
In addition to above, there has to be a check for every `vagrant service-manager env docker` call, to make sure the certs are valid on the host before printing the information.

if (certs are copied at the time of their generation) + (invoked `env docker` call) + (certs at host and guest unmatch) -- > Copy certs

In case certs unmatch, the `env docker` call will copy the certs.

Having the mismatch of certs at host and guest is a possibility, we should do sanity check before showing the information. Related issue https://github.com/projectatomic/vagrant-service-manager/issues/80

hferentschik:
Personally I think, all that's needed is a `regenerate-certs` command which the user can trigger in this case. The error is pretty clear, if the certificates don't match. Also, if the certs are copied during provisioning and the user always uses Vagrant commands to interact with the VM, a certificate mismatch should not occur. I thought there was already an issue for `regenerate-certs`, but I cannot find it right now.

navidshaikh:
> Personally I think, all that's needed is a `regenerate-certs` command which the user can trigger in this case. 

I think it is responsibility of  vsm to ensure the information provided is correct and the certs are valid and it should be a default check, the `copy` of certs will happen only if the there is a mismatch with a proper message. Also, we can delegate the option `env docker --re-download`  to user where user can explicitly ask for re downloading the certs from the box, however there should be a check by default for every run of `env docker`.

@bexelbie : thoughts?

bexelbie:
As long as the check does not overly slow down the `env` call then it is fine to recheck.  I agree that we need `--re-download` (not sure about the name).  However, we also need a `regenerate-certs` function as well.

brgnepal:
@bexelbie @navidshaikh  Please correct me if I am wrong.

Flag `--re-download(or any other good name)` will only\* download the certs from box while `regenerate-certs` command will only\* re-generate certs in the VM not copy it to host.

If we want both function i.e. re-generate certs and automatically copy the newly generate certs into host, then  we need proper naming here. Both the terms are looking different for me and doing different purpose IMO.

navidshaikh:
> As long as the check does not overly slow down the env call then it is fine to recheck.

I would rather have the call slow than showing up information without verification.

> However, we also need a regenerate-certs function as well.

yes. +1

PR #126, opened 2016-03-24
Fix-122: Certs copied at the time of generation

brgnepal:
Fix https://github.com/projectatomic/vagrant-service-manager/issues/122
- Implemented based on [`this comment`](https://github.com/projectatomic/vagrant-service-manager/issues/122#issuecomment-199314448)
- Extracted copying the certs operation another file so that it can be used from anywhere

My logs - https://gist.github.com/budhrg/523b8f87c954aefff9c8

### DIFF
--- a/lib/vagrant-service-manager/plugin_util.rb
+++ b/lib/vagrant-service-manager/plugin_util.rb
@@ -1,0 +1,22 @@
+module Vagrant
+  module ServiceManager
+    module PluginUtil
+      def self.copy_certs_to_host(machine, path, ui, commented_message = false)
+        Dir.mkdir(path) unless Dir.exist?(path)
+
+        # Copy the required client side certs from inside the box to host machine
+        message = "Copying TLS certificates to #{path}"
+        message = '# ' + message.to_s if commented_message
+        ui.info(message)
+        machine.communicate.download("#{DOCKER_PATH}/ca.pem", path.to_s)
+        machine.communicate.download("#{DOCKER_PATH}/cert.pem", path.to_s)
+        machine.communicate.download("#{DOCKER_PATH}/key.pem", path.to_s)
+      end
+
+      def self.host_docker_path(machine)
+        # Path to the private_key and where we will store the TLS Certificates
+        File.expand_path("docker", machine.data_dir)
+      end
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix #122: Certs copied at the time of generation @budhrg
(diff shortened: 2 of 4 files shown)


lib/vagrant-service-manager/plugin_util.rb
```diff
@@ -0,0 +1,18 @@
+module Vagrant
+  module ServiceManager
+    module PluginUtil
+      def self.copy_certs_to_host(machine, path)
+        Dir.mkdir(path) unless Dir.exist?(path)
+
```
navidshaikh, line 6: We need to put the message `"# Copying TLS certificates to #{secrets_path}"` here.
This will serve two purposes:
- Display message during box boot up process
- Display message during `env docker`  call given the certs at host are invalid and we needed to copy

navidshaikh:
@budhrg : update the changelog as well please. :-)

brgnepal:
@navidshaikh Updated PR.

navidshaikh:
I tested this PR with scenarios where
 1 - the certs are copied to host when box boots up
 2 - the certs are copied to host with invocation of `env docker` when the certs do not match

works as expected.

The log message we are printing when we actually copy the certs starts with `#`. 
`# Copying TLS certificates to #{secrets_path}`
which is needed for scenario 2 mentioned above (for `eval` to work), however for scenario 1 above, we do not need the `#`.
Although this is just a cosmetic change, I logged it for reference and to improve upon.

navidshaikh:
The PR LGTM and ready to be merged.
@bexelbie : ping

brgnepal:
@navidshaikh Updated the PR to support scenarios where we need to log message 

```
# Copying TLS certificates to #{secrets_path}
```

during `... env docker` command.
On `vagrant up` flow it will print  as vagrant default log.

My terminal logs here - https://gist.github.com/budhrg/24fe634c8845ebbad762

navidshaikh:
> Updated the PR to support scenarios where we need to log message

:+1: 

> My terminal logs here - https://gist.github.com/budhrg/24fe634c8845ebbad762

You can just use the comment section here to paste the logs.

LGTM

hferentschik:
> # Copying TLS certificates to #{secrets_path}

This line should not be displayed as part of `vagrant service-manager env docker`. It is by no means needed for the working of this command and just adds confusion. It should only  be printed as part of the bootstrapping process and there indeed w/o the '#'.

navidshaikh:
> This line should not be displayed as part of vagrant service-manager env docker. It is by no means needed for the working of this command and just adds confusion.

The line will only be printed when command `vagrant service-manager env docker` is invoked and the certs mismatch. 

If command `env docker` is invoked and certs match, the line won't be printed, the command will just print the docker connection information.

brgnepal:
@navidshaikh Rebased PR.

praveenkumar:
LGTM

PR #140, opened 2016-04-01
Fix #133: Adds restart command for services

### LINKED ISSUE #133: Add restart command for services
Initially we need this for docker if the user wants to force a docker daemon restart.

Ideally we need this for all services.

navidshaikh:
Fixes #133 
- Example execution:
  $ vagrant service-manager restart docker
  $ vagrant service-manager restart openshift
  $ vagrant service-manager restart kubernetes
  $ vagrant service-manager restart k8s
  
  For OpenShift and Kubernetes services, plugin uses sccli utility to restart the service.
  For other services, plugin uses systemctl utility to restart the service.
  If service can not be restarted, it exits with code 126.

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -5,6 +5,7 @@ module Vagrant
   module ServiceManager
     DOCKER_PATH = '/home/vagrant/.docker'
     SUPPORTED_SERVICES = ['docker', 'openshift', 'kubernetes']
+    SCCLI_SERVICES = ["openshift", "k8s"]
 
     class Command < Vagrant.plugin(2, :command)
       OS_RELEASE_FILE = '/etc/os-release'
@@ -77,6 +78,16 @@ module Vagrant
           end
         when '--help', '-h'
           print_help
+        when "restart"
+          self.exit_if_machine_not_running
+          case subcommand
+          when '--help', '-h'
+            print_help(type: command)
+          else
+            restart_service(subcommand)
+          end
+        when "help"
+            self.print_help
         else
           print_help(exit_status: 1)
         end
@@ -253,6 +264,26 @@ module Vagrant
           end
         end
       end
+
+      def restart_service(service)
+
+        if SCCLI_SERVICES.include? service
+          # TODO : Handle the case where user wants to pass extra arguments to OpenShift service
+          command = "sudo sccli #{service}"
+        else
+          command = "sudo systemctl restart #{service}"
+        end
+
+        with_target_vms(nil, { single_target: true}) do |machine|
+          machine.communicate.execute(command) do |type, error|
+            if type == :stderr
+              @env.ui.error(error)
+              exit 126
+            end
+          end
+        end
+      end
+
     end
   end
 end
--- a/locales/en.yml
+++ b/locales/en.yml
@@ -24,6 +24,7 @@ en:
           Commands:
                env          displays connection information for services in the box
                box          displays version and release information for the box
+               restart      restarts the given systemd service in the box
 
           Options:
                -h, --help   print this help
@@ -52,6 +53,20 @@ en:
           Options:
                 --script-readable  display information in a script readable format
                 -h, --help         print this help
+
+        restart: |-
+          Restarts the service
+
+          Usage: vagrant service-manager restart <object> [options]
+
+          Object:
+              Object is a service provided via sccli or systemd. For example: docker, k8s, or openshift etc.
+
+          Options:
+                -h, --help         print this help
+
+          Examples:
+            vagrant service-manager restart docker
 
       env:
         docker:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix #133: Adds restart command for services @navidshaikh
 - Fix #152: Makes plugin backward compatible with docker 1.8.2 for docker version API @navidshaikh
 - Fix #150: Adds .gitattributes to fix the CHANGELOG.md merge conflicts @bexelbie
 - Fix #142: Removes # before human readable output of openshift env info @navidshaikh
@@ -11,6 +12,7 @@
 - Fix #89: Improve help output for service-manager -h @budhrg
 - Vagrant way of showing information using 'locale' @budhrg
 - cygwin eval hint now removes colors and env uses export @bexelbie
+- Fix #131: Fixes starting OpenShift service by default for CDK box @navidshaikh
 
 ## v0.0.5 Mar 29, 2016
 - Fix #127: vagrant-service-manager 0.0.5 release @navidshaikh


lib/vagrant-service-manager/command.rb
```diff
@@ -221,6 +229,28 @@ def print_vagrant_box_version(script_readable = false)
           end
         end
       end
+
+      def restart_service(service)
+        sccli_services = ["openshift", "kubernetes", "k8s"]
```
brgnepal: Probably you would like to extract this as `SCCLI_SERVICES` inside `ServiceManager` namespace(Module) like https://github.com/projectatomic/vagrant-service-manager/blob/master/lib/vagrant-service-manager/config.rb#L5.
This way it could be used in any other file if required. Name in all caps is naming convention for constant.

lib/vagrant-service-manager/command.rb
```diff
@@ -221,6 +230,28 @@ def print_vagrant_box_version(script_readable = false)
...
         end
       end
+
+      def restart_service(service)
+
+        if SCCLI_SERVICES.include? service
+          # since sccli understand k8s
+          service = "k8s" if service == "kubernetes"
+          # TODO : Handle the case where user wants to pass extra arguments to OpenShift service
+          command = "sudo sccli #{service}"
```
bexelbie, line 272: I don't think this works for restarting services.  I don't think `sccli` has that concept yet.  @praveenkumar ?

praveenkumar, line 272: @bexelbie yes it does work. Me and @LalatenduMohanty  tested out before iirc. I also checked with @navidshaikh and looks work without any issue (Not test for plugin point of view but on newly created box and manually executing those command).

brgnepal:
I think it would be better if we add some info or output message on successful completion of restart command. Currently, there is nothing and it feels like nothing happened. See below.

```
➜  vagrant-service-manager git:(fix-133) ✗ vssh -c "sudo systemctl status docker"
● docker.service - Docker Application Container Engine
   Loaded: loaded (/usr/lib/systemd/system/docker.service; enabled; vendor preset: disabled)
  Drop-In: /usr/lib/systemd/system/docker.service.d
           └─flannel.conf
   Active: active (running) since Mon 2016-04-04 08:29:06 EDT; 1min 57s ago
     Docs: http://docs.docker.com
.................

➜  vagrant-service-manager git:(fix-133) ✗ be vagrant service-manager restart docker
<Nothing here> ( Better if we have some message like "Docker service restarted.")

➜  vagrant-service-manager git:(fix-133) ✗ vssh -c "sudo systemctl status docker"   
● docker.service - Docker Application Container Engine
   Loaded: loaded (/usr/lib/systemd/system/docker.service; enabled; vendor preset: disabled)
  Drop-In: /usr/lib/systemd/system/docker.service.d
           └─flannel.conf
   Active: active (running) since Mon 2016-04-04 08:31:17 EDT; 4s ago
     Docs: http://docs.docker.com
........................
```

lib/vagrant-service-manager/command.rb
```diff
@@ -221,6 +230,28 @@ def print_vagrant_box_version(script_readable = false)
...
+        else
+          command = "sudo systemctl restart #{service}"
```
brgnepal, line 274: I think we should check the validity of service here. If service is not valid (suppose someone fire `vagrant service-manager restart abc`) then will raise error as `Invalid service`.

navidshaikh:
> I think it would be better if we add some info or output message on successful completion of restart command. Currently, there is nothing and it feels like nothing happened. See below.

I thought about adding the message as part of the successful execution of the command (when service is restarted successfully). However, I logged in the box and noticed the behavior of `systemctl` command. `systemctl restart <service>` does not print output if command executed successfully and I followed the same pattern for `restart`. In case of errors, the error messages are shown accordingly.

lib/vagrant-service-manager/command.rb
bexelbie, line 272: @praveenkumar good to know.  My read of `sccli` led me to believe that it only had starts and not restarts ...

lib/vagrant-service-manager/command.rb
navidshaikh, line 274: > I think we should check the validity of service here. If service is not valid (suppose someone fire vagrant service-manager restart abc) then will raise error as Invalid service.

Thats handled by `systemd` and `systemctl` show the correct error message which we relay via plugin. @budhrg : Try testing the PR with different scenarios and let me know if something is missing ?

bexelbie:
+1 until we have a user case for needing the message, omitting it, inline with existing Linux tools feels right.

brgnepal:
Tested it. LGTM.

navidshaikh:
This PR needs to update the new pattern in `--help` as we are following.

navidshaikh:
Updated the PR to incorporate the help for main help and subcommand help for `restart` command.

Here is the sample output of the help command

```
➜  vagrant-service-manager git:(fix-133) bundle exec vagrant service-manager -h   
Usage: vagrant service-manager <command> [options]

Commands:
     env          displays connection information for services in the box
     box          displays version and release information for the box
     restart      restarts the given systemd service in the box

Options:
     -h, --help   print this help

For help on any individual command run `vagrant service-manager COMMAND -h`

➜  vagrant-service-manager git:(fix-133) bundle exec vagrant service-manager restart -h
Restarts the given systemd service in the box

Usage: vagrant service-manager restart <object> [options]

Object:
    Object can be any available service configured in the box using systemd unit files

Options:
      -h, --help         print this help

Examples:
  vagrant service-manager restart docker
  vagrant service-manager restart kubernetes
  vagrant service-manager restart openshift
```

brgnepal:
@navidshaikh @bexelbie 
I think one example is enough to show usage.

locales/en.yml
```diff
@@ -53,6 +54,22 @@ en:
                 --script-readable  display information in a script readable format
                 -h, --help         print this help
 
+        restart: |-
+          Restarts the given systemd service in the box
```
bexelbie: Restarts the service

No need to mention systemd or box

locales/en.yml
```diff
@@ -53,6 +54,22 @@ en:
...
+
+          Usage: vagrant service-manager restart <object> [options]
+
+          Object:
+              Object can be any available service configured in the box using systemd unit files
+
```
bexelbie, line 64: Object is a service provided via sccli or systemd.  For example:
docker, kubernetes (or k8s), or openshift.

locales/en.yml
```diff
@@ -53,6 +54,22 @@ en:
...
+          Options:
+                -h, --help         print this help
+
+          Examples:
+            vagrant service-manager restart docker
+            vagrant service-manager restart kubernetes
+            vagrant service-manager restart openshift
```
bexelbie: I think the last two examples can be deleted if the text above is used.

bexelbie:
With consideration of the changes above .. LGTM

locales/en.yml
brgnepal, line 64: > docker, kubernetes (or k8s), or openshift.

Does this limits to only three services? Right now it looks like `sccli` or `systemd` services are only `docker, kubernetes or openshift`. 
IMO it would be better if we re-write it as `docker, kubernetes (or k8s), openshift etc`. This gives hint to user that you can user more `systemd` services too like `httpd` etc.

What do you think @bexelbie?

navidshaikh:
I've updated the help text with suggestions and rebased the PR as well.

lib/vagrant-service-manager/command.rb
LalatenduMohanty, line 272: @bexelbie Basically it is not very clean now and we should clean it up in future. At this point of time https://github.com/projectatomic/adb-utils/blob/master/utils/sccli.sh#L9 does the trick

lib/vagrant-service-manager/command.rb
```diff
@@ -5,6 +5,7 @@ module Vagrant
   module ServiceManager
     DOCKER_PATH = '/home/vagrant/.docker'
     SUPPORTED_SERVICES = ['docker', 'openshift', 'kubernetes']
+    SCCLI_SERVICES = ["openshift", "kubernetes", "k8s"]
```
LalatenduMohanty: This looks wrong. It should be ["openshift",  "k8s"] . Looks like it is done only to make code logic easy?

LalatenduMohanty: Looking at the time constraint here, my comment should not stop you from merging the patch. We can re-factor the code post release.

navidshaikh: > Looks like it is done only to make code logic easy?

Although `sccli` understands only `openshift` and `k8s`, the plugin provides an extra option for users to either enter `k8s` or `kubernetes` as a choice of service name to restart.

LalatenduMohanty: If we think Kubernetes and k8s both are required then we should fix sccli to include Kubernetes and then use it here. That way we would be consistent across the board.

navidshaikh: Unless we have fix for it in `sccli` lets carry the patch in VSM, later we remove it.

LalatenduMohanty: @navidshaikh you mean we will remove kubernetes? or k8s?

lib/vagrant-service-manager/command.rb
```diff
@@ -253,6 +264,28 @@ def print_vagrant_box_version(script_readable = false)
           end
         end
       end
+
+      def restart_service(service)
+
+        if SCCLI_SERVICES.include? service
+          # since sccli understand k8s
+          service = "k8s" if service == "kubernetes"
```
navidshaikh: @LalatenduMohanty : This line will be removed after we have support for `k8s` and `kubernetes` both in `sccli`. Here we are accepting both the terms for kubernetes service.

LalatenduMohanty:
LGTM

Issue #141, opened 2016-04-04
Improve output of 'vagrant service-manager env'


hferentschik:
Currently we have:

```
$ vagrant service-manager env
# Showing the status of providers in the vagrant box:

# Set the following environment variables to enable access to the
# docker daemon running inside of the vagrant virtual machine:
export DOCKER_HOST=tcp://10.1.2.2:2376
export DOCKER_CERT_PATH=/Users/hardy/work/developer-platform/openshift-vagrant/cdk    -v2/.vagrant/machines/cdk/virtualbox/docker
export DOCKER_TLS_VERIFY=1
export DOCKER_API_VERSION=1.20
# run following command to configure your shell:
# eval "$(vagrant service-manager env docker)"


# You can access the OpenShift console on: https://10.1.2.2:8443/console
# To use OpenShift CLI, run: oc login https://10.1.2.2:8443
```

To be honest, I have no idea what a user should do with this information. There is no indication what comes from which service. Also, the amount of 'comments' (#) is confusing. Last but not least, the title "Showing the status of providers" is wrong. The output atm does not reflect ay service status.

IMO, 'vagrant service-manager env' should just display the configured services and maybe their status. Something along these lines:

```
Configured services:
Docker     - running 
OpenShift  - stopped
```

If not something like this, I'd rather have this command just display the help info. As said, concatenating the output of the various services is not helpful.

navidshaikh:
Idea here with `vagrant service-manager env` command is to show the status of all the active providers in the box along with their connection information. I agree that the output just concatenate connection information from different providers, however it could have been more elaborative with right descriptive messages OR more user friendly output.

> IMO, 'vagrant service-manager env' should just display the configured services and maybe their status. Something along these lines:
> 
> Configured services:
> Docker     - running 
> OpenShift  - stopped

+1 , IMO we should start the output of the `vagrant service-manager env` command with above summary and then display the connection information followed by summary.

navidshaikh:
Here is the sample output using the PR created

```
$vagrant service-manager env
Configured services:
docker - running
openshift - running
kubernetes - stopped


# Set the following environment variables to enable access to the
# docker daemon running inside of the vagrant virtual machine:
export DOCKER_HOST=tcp://172.28.128.4:2376
export DOCKER_CERT_PATH=/home/nshaikh/work/src/vagrant-service-manager/.vagrant/machines/default/virtualbox/docker
export DOCKER_TLS_VERIFY=1
[...]

PR #143, opened 2016-04-04
Fix#75 #141: Improves `vagrant service-manager env` output

### LINKED ISSUE #75: Exit with 0 if provider is not available for command `vagrant service-manger env`
If user has executed `vagrant service-manger env` for printing all active providers in the vagrant box and say a provider (OpenShift) is not running, the command should exit with zero and do not print any error message. The goal of the command is to display the active available providers in the box.

navidshaikh:
Fixes #75 and #141 
- Adds summary of the configured services and their running status
- Does not exit with non-zero code if one of the service is stopped
- Adds separation between different provider connection info

Sample output:

```
$ vagrant service-manager env
Configured services:
docker - running
openshift - running
kubernetes - stopped

docker env:
export DOCKER_TLS_VERIFY=1

openshift env:
# You can access the OpenShift console on: https://172.28.128.4:8443/console
# To use OpenShift CLI, run: oc login https://172.28.128.4:8443

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix #75 and #141: Improves `vagrant service-manager env` output @navidshaikh
 - Fix#146: Updates docker 1.9.1 API call for `docker version` @navidshaikh
 - Update IP detection routine and fix for libvirt @bexelbie
 - Fix #50: Add --help @budhrg
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -4,6 +4,7 @@ require 'digest'
 module Vagrant
   module ServiceManager
     DOCKER_PATH = '/home/vagrant/.docker'
+    SUPPORTED_SERVICES = ['docker', 'openshift', 'kubernetes']
 
     class Command < Vagrant.plugin(2, :command)
       OS_RELEASE_FILE = '/etc/os-release'
@@ -98,8 +99,22 @@ module Vagrant
 
       def print_all_provider_info
         @env.ui.info I18n.t('servicemanager.commands.env.nil')
-        self.execute_docker_info
-        self.execute_openshift_info
+
+        running_services = []
+        SUPPORTED_SERVICES.each do |service|
+          status = if check_if_a_service_is_running?(service)
+                     running_services << service
+                     I18n.t('servicemanager.commands.env.status.running')
+                    else
+                     I18n.t('servicemanager.commands.env.status.stopped')
+                    end
+          @env.ui.info("#{service} - #{status}")
+        end
+
+        running_services.each do |e|
+          @env.ui.info("\n#{e} env:")
+          public_send("execute_#{e}_info")
+        end
       end
 
       def execute_openshift_info(script_readable = false)
--- a/locales/en.yml
+++ b/locales/en.yml
@@ -81,4 +81,7 @@ en:
         service_not_running: |-
           # %{name} service is not running in the vagrant box.
         nil: |-
-          # Showing the status of providers in the vagrant box:
+          Configured services:
+        status:
+          running: running
+          stopped: stopped


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix #75 and #141: Improves `vagrant service-manager env` output
```
brgnepal: Forgot to add your username.

lib/vagrant-service-manager/command.rb
```diff
@@ -86,8 +87,20 @@ def check_if_a_service_is_running?(service)
 
       def print_all_provider_info
         @env.ui.info I18n.t('servicemanager.commands.env.nil')
-        self.execute_docker_info
-        self.execute_openshift_info
+
+        running_services = []
+        SUPPORTED_SERVICES.each do |service|
+          if self.check_if_a_service_is_running?(service)
```
brgnepal: `self` is unnecessary here.

navidshaikh: updated the PR.

lib/vagrant-service-manager/command.rb
```diff
@@ -86,8 +87,20 @@ def check_if_a_service_is_running?(service)
 
       def print_all_provider_info
         @env.ui.info I18n.t('servicemanager.commands.env.nil')
-        self.execute_docker_info
-        self.execute_openshift_info
+
+        running_services = []
+        SUPPORTED_SERVICES.each do |service|
```
brgnepal, line 104: You may want to change these `each` blocks as:

```
        SUPPORTED_SERVICES.each do |service|
          status = if check_if_a_service_is_running?(service)
                     running_services << service
                     'running'
                   else
                     'stopped'
                   end
          @env.ui.info("#{service.capitalize} - #{status}")
        end
        running_services.each do |service|
          @env.ui.info("\n#{service.capitalize} env:")
          public_send("execute_#{service}_info")
        end
```
- Capitalization of service name
- One time usage of `@env.ui.info` 

Rest LGTM.

navidshaikh, line 104: > Capitalization of service name

`openshift` will be capitalized as `Openshift`, which ideally should be `OpenShift`. Instead of that here the service name is matching with the systemd name of the service.

> One time usage of @env.ui.info

Although `@env.ui.info` is written twice, but executed once per service in loop iteration and we are adding one more variable `status` in your suggestion! :-)

brgnepal, line 104: > openshift will be capitalized as Openshift, which ideally should be OpenShift. Instead of that here the service name is matching with the systemd name of the service.

Would agree if we want to keep service name as that of `systemd` name.

> Although @env.ui.info is written twice, but executed once per service in loop iteration and we are adding one more variable status in your suggestion! :-)

Writing `@env.ui.info` once is more salable even if we trades off it with one more variable.

navidshaikh:
> Should we allow localization of "running" and "stopped" ?

@bexelbie : Added strings in `locales/en.yml`

lib/vagrant-service-manager/command.rb
```diff
@@ -86,8 +87,22 @@ def check_if_a_service_is_running?(service)
...
+          status = if check_if_a_service_is_running?(service)
+                     running_services << service
+                     I18n.t('servicemanager.commands.env.status.running')
```
brgnepal, line 107: :+1: for localization.

lib/vagrant-service-manager/command.rb
navidshaikh, line 104: >  Would agree if we want to keep service name as that of systemd name.

I kept the service name in the command output as that of respective systemd service name.

> Writing @env.ui.info once is more salable even if we trades off it with one more variable.

Updated the PR accordingly.

locales/en.yml
```diff
@@ -68,4 +68,9 @@ en:
         service_not_running: |-
           # %{name} service is not running in the vagrant box.
         nil: |-
-          # Showing the status of providers in the vagrant box:
+          Configured services:
+        status:
+          running: |-
+            running
```
brgnepal: I believe you can also write it as `running: "running"`. `|-` or simply `|` is used for multi-line string.

navidshaikh: @budhrg : updated the PR :-)

locales/en.yml
```diff
@@ -68,4 +68,9 @@ en:
         service_not_running: |-
           # %{name} service is not running in the vagrant box.
         nil: |-
-          # Showing the status of providers in the vagrant box:
+          Configured services:
+        status:
+          running:
```
brgnepal: This could be `running: "running"` as `|- (or | )` is used for multi-line string.

bexelbie:
LGTM

PR #145, opened 2016-04-05
cygwin eval hint removes colors & env uses export

### LINKED ISSUE #10: Print unix format of commands under cygwin

### LINKED ISSUE #112: service-manager prints unwanted characters in bash on windows
On windoes 10, in bash from cygwin:
$ eval "$(vagrant service-manager env docker)"
-bash: $'\E[0m#': command not found
-bash: $'\r': command not found
-bash: $'\E[0m\r': command not found

I think that @env.ui.info() (or something even deeper in ruby) is responsible for this. I tried to add puts() and print() around "message" and it still printed these characters. Simple @env.ui.info("hello") adds them too. Also more @env.ui.info() invocation insterts more such characters.

bexelbie:
Fix #112
Fix #10
Replace PR #100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Vagrant way of showing information using 'locale' @budhrg
+- cygwin eval hint now removes colors and env uses export @bexelbie
 
 ## v0.0.5 Mar 29, 2016
 - Fix #127: vagrant-service-manager 0.0.5 release @navidshaikh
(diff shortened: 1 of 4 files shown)


bexelbie:
@optak can you please test - I do not have an cygwin environment.

navidshaikh:
@optak tested the PR, see https://github.com/projectatomic/vagrant-service-manager/issues/112#issuecomment-205808376 

LGTM!

PR #147, opened 2016-04-05
Fix#146: Update docker 1.9.1 API call for `docker version`

### LINKED ISSUE #146: Update docker 1.9.1 API call for  `docker version`
For docker connection information, we are showing `DOCKER_API_VERSION=1.20` in the output of `env docker` command.

ATM, to provide this info in output, we are consuming docker API as `docker version --format '{{.Server.ApiVersion}}'`

Although output of from docker 1.8.2 to docker 1.9.1 for command `docker version` has not changed, the template needs to be updated to fetch the right field's value from the output.

The updated template to be used with 1.9.1 should be `docker version --format '{{.Server.APIVersion}}'`

navidshaikh:
Fixes #146

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix#146: Updates docker 1.9.1 API call for `docker version` @navidshaikh
 - Update IP detection routine and fix for libvirt @bexelbie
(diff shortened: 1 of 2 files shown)


bexelbie:
LGTM

brgnepal:
Tested and working as expected.

PR #148, opened 2016-04-05
Update IP detection routine

bexelbie:
IP detection was failing for libvirt when more than 1 nic was
configured.  This patch makes detection the same for all hosts.

The detection selects the last nic's IPv4 Address from all nics
that are up.

The custom detection for VirtualBox is no longer needed as we now
force add a second nic.

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -135,14 +135,7 @@ module Vagrant
       def find_machine_ip
         with_target_vms(nil, {:single_target=>true}) do |machine|
           # Find the guest IP
-          if machine.provider_name == :virtualbox then
-            # VirtualBox automatically provisions an eth0 interface that is a NAT interface
-            # We need a routeable IP address, which will therefore be found on eth1
-            command = "ip addr show eth1 | awk 'NR==3 {print $2}' | cut -f1 -d\/"
-          else
-            # For all other provisions, find the default route
-            command = "ip route get 8.8.8.8 | awk 'NR==1 {print $NF}'"
-          end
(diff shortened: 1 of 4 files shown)


bexelbie:
Companion PR in adb-utils https://github.com/projectatomic/adb-utils/pull/97

lib/vagrant-service-manager/command.rb
```diff
@@ -123,14 +123,7 @@ def print_openshift_info(url, console_url, script_readable = false)
...
           # Find the guest IP
-          if machine.provider_name == :virtualbox then
-            # VirtualBox automatically provisions an eth0 interface that is a NAT interface
-            # We need a routeable IP address, which will therefore be found on eth1
-            command = "ip addr show eth1 | awk 'NR==3 {print $2}' | cut -f1 -d\/"
-          else
-            # For all other provisions, find the default route
-            command = "ip route get 8.8.8.8 | awk 'NR==1 {print $NF}'"
-          end
+          command = "ip -o -4 addr show up |egrep -v ': docker|: lo' |tail -1 | awk '{print $4}' |cut -f1 -d\/"
```
navidshaikh, line 138: I liked the idea and fix, we don't really need the provider based IP detection since we add the private networking for required providers.

brgnepal:
Tested and found correct IP. LGTM.

```
$ bundle exec vagrant service-manager env docker
export DOCKER_CERT_PATH=/home/budhram/redhat/vagrant-service-manager/.vagrant/machines/default/virtualbox/docker
export DOCKER_TLS_VERIFY=1
```

navidshaikh:
Thanks @budhrg  for testing it. LGTM.

lib/vagrant-service-manager/command.rb
hferentschik, line 138: I still believe there must be another way of doing this ;-) What if the users adds additional network interfaces? But let's go with this for now.

Issue #69, opened 2016-02-24
Create a guide about how to add new services to the plugin


LalatenduMohanty:
We need to create a guide about how to add new services to the plugin. So that it is easy for other contributors.

brgnepal:
@LalatenduMohanty I hope it is developers/contributors point of view.

bexelbie:
Should this be delayed pending #135

bexelbie:
@budhrg I am looking to you to provide the core code examples, per your comments in #155 then @Preeticp or I can take over :)

brgnepal:
@bexelbie : When you say code examples, should I put it in some README with few instructions like:
- Create service class like below:
  ....
- Implement these methods as below etc 

Could you give me rough expected outline?

bexelbie:
@budhrg Exactly

navidshaikh:
@budhrg : ping

PR #155, opened 2016-04-07
Fix #135: Refactor command.rb to make commands easier to add/maintain

### LINKED ISSUE #135: Refactor command.rb to make commands easier to add/maintain
Refact the code in command.rb should be refactored so that the service specific bits live somewhere like lib/vagrant-service-manager/services/docker.rb. Other code like find_machine_ip can go in a common library.

brgnepal:
Fix #135: Refactor command.rb
- Migrated code to respective services like status, showing info
- Move machine related code to guest capabilities
- Migrated common code into `plugin_util` file

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -1,5 +1,6 @@
 require_relative 'os'
 require 'digest'
+require_relative 'plugin_util'
 
 module Vagrant
   module ServiceManager
@@ -8,14 +9,15 @@ module Vagrant
     SCCLI_SERVICES = ['openshift', 'k8s']
     KUBE_NAMES = ['kubernetes', 'k8s']
     KUBE_SERVICES = [
-      "etcd",
-      "kube-apiserver",
-      "kube-controller-manager",
-      "kube-scheduler",
-      "kubelet",
-      "kube-proxy",
-      "docker"
+      'etcd', 'kube-apiserver', 'kube-controller-manager', 'kube-scheduler',
+      'kubelet', 'kube-proxy', 'docker'
     ]
+    # NOTE: SERVICES_MAP[<service>] will give fully-qualified service class name
+    # Eg: SERVICES_MAP['docker'] gives Vagrant::ServiceManager::Docker
+    SERVICES_MAP = {
+      'docker' => Docker, 'openshift' => OpenShift,
+      'kubernetes' => Kubernetes
+    }
 
     class Command < Vagrant.plugin(2, :command)
       OS_RELEASE_FILE = '/etc/os-release'
@@ -26,7 +28,7 @@ module Vagrant
 
       def exit_if_machine_not_running
         # Exit from plugin with status 3 if machine is not running
-        with_target_vms(nil, {:single_target=>true}) do |machine|
+        with_target_vms(nil, single_target: true) do |machine|
           if machine.state.id != :running
             @env.ui.error I18n.t('servicemanager.machine_should_running')
             exit 3
@@ -41,23 +43,12 @@ module Vagrant
         when 'env'
           exit_if_machine_not_running
           case subcommand
-          when 'docker'
+          when 'docker', 'openshift'
             case option
             when nil
-              execute_docker_info
+              execute_service(subcommand)
             when '--script-readable'
-              execute_docker_info(true)
-            when '--help', '-h'
-              print_help(type: command)
-            else
-              print_help(type: command, exit_status: 1)
-            end
-          when 'openshift'
-            case option
-            when nil
-              execute_openshift_info
-            when '--script-readable'
-              execute_openshift_info(true)
+              execute_service(subcommand, script_readable: true)
             when '--help', '-h'
               print_help(type: command)
             else
@@ -67,7 +58,7 @@ module Vagrant
             # display information about all the providers inside ADB/CDK
             print_all_provider_info
           when '--script-readable'
-            print_all_provider_info(true)
+            print_all_provider_info(script_readable: true)
           when '--help', '-h'
             print_help(type: command)
           else
@@ -92,6 +83,8 @@ module Vagrant
               print_vagrant_box_version
             when '--script-readable'
               print_vagrant_box_version(true)
+            when '--help', '-h'
+              print_help(type: command)
             else
               print_help(type: command, exit_status: 1)
             end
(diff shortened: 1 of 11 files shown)


lib/vagrant-service-manager/command.rb
```diff
@@ -34,7 +34,7 @@ def execute
           when 'docker'
```
bexelbie: Can you combine the docker and openshift subcommands with a temporary extra check around --script-readable for openshift ... it'd be nice to get a follow up PR that removes that.

lib/vagrant-service-manager/command.rb
```diff
@@ -76,23 +78,36 @@ def execute
...
+        when 'help'
+          print_help
         else
           print_help(exit_status: 1)
         end
       end
 
+      def execute_service(name, options = {})
+        with_target_vms(nil, single_target: true) do |machine|
+          klass_name = case name
```
bexelbie: why klass and not class?

bexelbie:
@budhrg A few comments inline.

I am so grateful to see this PR - it is fantastic.

I haven't tested it, and will do so when I am both on better internet and it has been rebased.  But it LGTM other than above so far.

lib/vagrant-service-manager/plugin_util.rb
```diff
@@ -17,6 +17,37 @@ def self.host_docker_path(machine)
...
+        Dir[path + "/*"].each do |f|
+          guest_file_path = "#{DOCKER_PATH}/#{File.basename(f)}"
+          guest_sha = machine.guest.capability(:sha_id, guest_file_path)
+          return false if sha_id(File.read(f)) != guest_sha
+        end
+
+        true
+      end
+
+      def self.service_running?(machine, service)
```
bexelbie, line 72: Should the method of determining if a service is running be in the service file? i.e. in lib/vagrant-service-manager/services/open_shift.rb ?

bexelbie:
As a part of this can we rename "lib/vagrant-service-manager/services/open_shift.rb" to lib/vagrant-service-manager/services/openshift.rb"

lib/vagrant-service-manager/command.rb
brgnepal: `class` is keyword :)

lib/vagrant-service-manager/plugin_util.rb
brgnepal, line 72: @bexelbie That make sense to me :smile: 
This way, each service will able to determine whether they are running or not by themselves.

lib/vagrant-service-manager/command.rb
brgnepal: @bexelbie I have way to refactor `execute` method content. It will reduce to only one or two lines instead of its never ending content.
On my way to it :smile:

brgnepal:
> As a part of this can we rename "lib/vagrant-service-manager/services/open_shift.rb" to lib/vagrant-service-manager/services/openshift.rb"

@bexelbie I think I have kept it for some purpose intentionally. Let me give you a hint of my intention.
Suppose I want to get all the services class name like `Docker`, `OpenShift`, etc.
Now, if i have file name as `openshift`, I cannot able to get class name as `OpenShift` however, if I use file name as `open_shift`, I know that I have to split this file name base one `_`(underscore) and get two words which I capitalize each of first letter and then join to get the required class name.

This helps us to leverage the dynamic nature of Ruby. :smile:

budhrg:
@navidshaikh @bexelbie Need more work this PR.

brgnepal:
@bexelbie @navidshaikh : Updated and rebased PR.

Few of command outputs:

```
$ vagrant service-manager env                                                
docker env:
export DOCKER_HOST=tcp://10.1.2.2:2376
export DOCKER_CERT_PATH=/home/budhram/redhat/vagrant-service-manager/.vagrant/machines/default/virtualbox/docker
export DOCKER_TLS_VERIFY=1

openshift env:
# You can access the OpenShift console on: https://10.1.2.2:8443/console
# To use OpenShift CLI, run: oc login https://10.1.2.2:8443

$ vagrant service-manager env --script-readable
DOCKER_HOST=tcp://10.1.2.2:2376
DOCKER_CERT_PATH='/home/budhram/redhat/vagrant-service-manager/.vagrant/machines/default/virtualbox/docker'
DOCKER_TLS_VERIFY=1
DOCKER_API_VERSION=1.21
OPENSHIFT_URL=https://10.1.2.2:8443
OPENSHIFT_WEB_CONSOLE=https://10.1.2.2:8443/console
```

lib/vagrant-service-manager/command.rb
bexelbie: Can this be `when 'docker', 'openshift'` ?

bexelbie:
I added one line comment: https://github.com/projectatomic/vagrant-service-manager/pull/155/files#r59863404

Also, I feel like this refactor is missing three things (and yes, the PR will be big, but it is a refactor):
1. We should have the service level files provide a status function where they can determine if they are running and tell us.  This affects lines like this: https://github.com/projectatomic/vagrant-service-manager/blob/master/lib/vagrant-service-manager/command.rb#L145
2. We should have a kube file created that parallels openshift and docker classes.  I know kube doesn't have all the features, but then it can have stubs.  See lines like this: https://github.com/projectatomic/vagrant-service-manager/blob/master/lib/vagrant-service-manager/command.rb#L163
3. We should have a generic example class file that we can use as a foundation for docs on how to add support for new services.

brgnepal:
@bexelbie, @navidshaikh : We could add 1st and 2nd suggestions as part of this PR but suggest to keep 3rd as part of fix for #69 since it's a new addition of changes rather than part of refactoring.

lib/vagrant-service-manager/command.rb
brgnepal: Yes @bexelbie . Thanks :+1:

brgnepal:
@bexelbie , @navidshaikh : Updated PR with bex's 1st and 2nd suggestions.

bexelbie:
I feel like lines 156+ of command.rb (where we special case on Kube) should still be cleaned up.  However, I am open to re-reading this with more sleep under my belt if there is a good reason.

brgnepal:
> I feel like lines 156+ of command.rb (where we special case on Kube) should still be cleaned up. 

@bexelbie , I think it is not needed ATM since the condition for kube will go once we start supporting showing Kube information. Then, it is matter of deleting those Kube related conditions.

navidshaikh:
@budhrg : Lets rebase this with master.

brgnepal:
@navidshaikh : Rebased with upstream.

lib/vagrant-service-manager/command.rb
```diff
@@ -8,14 +9,15 @@ module ServiceManager
...
-      "etcd",
-      "kube-apiserver",
-      "kube-controller-manager",
-      "kube-scheduler",
-      "kubelet",
-      "kube-proxy",
-      "docker"
+      'etcd', 'kube-apiserver', 'kube-controller-manager', 'kube-scheduler',
+      'kubelet', 'kube-proxy', 'docker'
     ]
```
LalatenduMohanty, line 14: I kind of like this format of listing services. Some coding standard prefer this as this is more readable.  Is there any reason we moved away from it?

brgnepal, line 14: @LalatenduMohanty : Well, i thought it would be good to group based on master and minion services which would be more readable in context of Kubernetes along with less lines.
Also, Ruby [multi-lines Array guidelines](https://github.com/bbatsov/ruby-style-guide#align-multiline-arrays) suggests this format.

navidshaikh:
LGTM

bexelbie:
LGTM

PR #156, opened 2016-04-07
 Fix #152: Makes plugin backward compatible with docker 1.8.2

### LINKED ISSUE #152: Make plugin backward compatible with docker 1.8.2 for docker version API
If user is running box with docker 1.8.2 , the plugin should be backward compatible to call the older API as well.

Related https://github.com/projectatomic/vagrant-service-manager/issues/146

navidshaikh:
Fixes #152
 Makes plugin backward compatible with docker 1.8.2 for docker version API

 With this PR, plugin can work with boxes having either docker 1.9.1 or 1.8.2 version

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -191,9 +191,14 @@ module Vagrant
           end
 
           api_version = ""
-          docker_apiversion = "docker version --format '{{.Server.APIVersion}}'"
-          machine.communicate.execute(docker_apiversion) do |type, data|
-            api_version << data.chomp if type == :stdout
+          docker_api_version = "docker version --format '{{.Server.APIVersion}}'"
+          unless machine.communicate.test(docker_api_version)
+            # fix for issue #152: Fallback to older Docker version (< 1.9.1)
+            docker_api_version.gsub!(/APIVersion/, 'ApiVersion')
+          end
+
+          machine.communicate.execute(docker_api_version) do |type, data|
+            api_version << data.chomp if type ==:stdout
           end
 
           # display the information, irrespective of the copy operation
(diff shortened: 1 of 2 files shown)


lib/vagrant-service-manager/command.rb
```diff
@@ -192,8 +192,16 @@ def execute_docker_info
 
           api_version = ""
           docker_apiversion = "docker version --format '{{.Server.APIVersion}}'"
```
brgnepal: `docker_api_version` is preferred name here.

lib/vagrant-service-manager/command.rb
```diff
@@ -192,8 +192,16 @@ def execute_docker_info
...
+            machine.communicate.execute(docker_apiversion) do |type, data|
+              api_version << data.chomp if type == :stdout
+            end
+          else
+            # fix for issue #152
+            docker_apiversion = "docker version --format '{{.Server.ApiVersion}}'"
+            machine.communicate.execute(docker_apiversion) do |type, data|
+              api_version << data.chomp if type ==:stdout
+            end
           end
```
brgnepal, line 202: Probably you would like to get the `docker_api_version` based on `test` condition and later execute. Then, `execute` block will come once. In such case, you may follow this pattern:

```
docker_api_version = "docker version --format '{{.Server.APIVersion}}'"
unless machine.communicate.test(docker_api_version)
    # Fallback to older Docker version (< 1.9.1)
    docker_api_version.gsub!(/APIVersion/, 'ApiVersion')            
end

machine.communicate.execute(docker_api_version) do |type, data|
  api_version << data.chomp if type ==:stdout
end
```

brgnepal:
LGTM apart from needed code refactoring.

LalatenduMohanty:
LGTM. @navidshaikh tested it for docker 1.9.1 and 1.8.2

Issue #11, opened 2016-02-09
Display Environment Variable for Kubernetes for use with kubectl


navidshaikh:
_From @bexelbie on December 18, 2015 4:32_

_Copied from original issue: projectatomic/vagrant-adbinfo#58_

bexelbie:
https://github.com/projectatomic/adb-atomic-developer-bundle/issues/250

PR #176, opened 2016-04-12
Fix#173: Shows if kubernetes services is running in the box

### LINKED ISSUE #173: `vagrant service-manager env` does not show the status of kube
Even if kube service is running inside the box, `vagrant service-manager env` does not show the service running, rather it shows it as `stopped`.

```
➜  vagrant-service-manager ✗ vagrant service-manager env
Configured services:
docker - running
openshift - stopped
kubernetes - stopped

➜  vagrant-service-manager git:(budhrg-fix-160) ✗ vagrant ssh
[vagrant@rhel-cdk ~]$ kubectl get nodes
NAME        LABELS                             STATUS    AGE
127.0.0.1   kubernetes.io/hostname=127.0.0.1   Ready     15m
```

navidshaikh:
Fixes #173

```
 $ vagrant service-manager env
 Configured services:
 docker - running
 openshift - stopped
 kubernetes - running

 docker env:
 export DOCKER_HOST=tcp://172.28.128.8:2376
 export DOCKER_CERT_PATH=/foo/bar/.vagrant/machines/default/virtualbox/docker
 export DOCKER_TLS_VERIFY=1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 - Add --script-readable to env and env docker @bexelbie
 - Fix #178: Add status command and separate status from env @bexelbie
+- Fix#173: Shows if kubernetes services is running in the box @navidshaikh
 
 ## v1.0.1 Apr 12, 2016
 - Updated SPEC (v1.0.0) for url, date and format @budhrg
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -6,6 +6,16 @@ module Vagrant
     DOCKER_PATH = '/home/vagrant/.docker'
     SUPPORTED_SERVICES = ['docker', 'openshift', 'kubernetes']
     SCCLI_SERVICES = ['openshift', 'k8s']
+    KUBE_NAMES = ['kubernetes', 'k8s']
+    KUBE_SERVICES = [
+      "etcd",
+      "kube-apiserver",
+      "kube-controller-manager",
+      "kube-scheduler",
+      "kubelet",
+      "kube-proxy",
+      "docker"
+    ]
 
     class Command < Vagrant.plugin(2, :command)
       OS_RELEASE_FILE = '/etc/os-release'
@@ -115,6 +125,7 @@ module Vagrant
       end
 
       def service_running?(service)
+        return kube_running? if KUBE_NAMES.include? service
         command = "systemctl status #{service}"
         with_target_vms(nil, {:single_target=>true}) do |machine|
           return machine.communicate.test(command)
@@ -139,6 +150,13 @@ module Vagrant
         end
       end
 
+      def kube_running?
+        KUBE_SERVICES.each do |service|
+          return false unless service_running?(service)
+        end
+        true
+      end
+
       def running_services
         running_services = []
         SUPPORTED_SERVICES.each do |service|
@@ -150,9 +168,10 @@ module Vagrant
       def print_all_provider_info(script_readable = false)
         running_services.each do |e|
           unless  script_readable
-            @env.ui.info("\n#{e} env:")
+            # since we do not have feature to show the kube connection information
+            @env.ui.info("\n#{e} env:") unless KUBE_NAMES.include? e
           end
-          public_send("execute_#{e}_info", script_readable)
+          public_send("execute_#{e}_info", script_readable) unless KUBE_NAMES.include? e
         end
       end
 


LalatenduMohanty:
Should we just check "kubectl get nodes" as Kubernetes setup is running as single node setup?

lib/vagrant-service-manager/command.rb
```diff
@@ -6,6 +6,15 @@ module ServiceManager
...
     SUPPORTED_SERVICES = ['docker', 'openshift', 'kubernetes']
     SCCLI_SERVICES = ['openshift', 'k8s']
+    KUBE_SERVICES = [
+      "etcd",
+      "kube-apiserver",
+      "kube-controller-manager",
+      "kube-scheduler",
+      "kubelet",
+      "kube-proxy",
+      "docker"
```
brgnepal, line 17: I think 'docker' is not required as per my discussion with @praveenkumar.

lib/vagrant-service-manager/command.rb
```diff
@@ -102,12 +111,20 @@ def print_help(config = {})
       end
 
       def check_if_a_service_is_running?(service)
+        return check_if_kube_is_running? if ['kubernetes', 'k8s'].include? service
```
brgnepal: You may want to extract `['kubernetes', 'k8s']`  as `KUBE_NAMES` or any preferred named constant in plugin level.

lib/vagrant-service-manager/command.rb
```diff
@@ -102,12 +111,20 @@ def print_help(config = {})
...
         command = "systemctl status #{service}"
         with_target_vms(nil, {:single_target=>true}) do |machine|
           return machine.communicate.test(command)
         end
       end
 
+      def check_if_kube_is_running?
+        KUBE_SERVICES.each do |service|
+          return false if not check_if_a_service_is_running?(service)
```
brgnepal: For such situation Ruby has `unless` keyword for `if not`.

lib/vagrant-service-manager/command.rb
```diff
@@ -102,12 +111,20 @@ def print_help(config = {})
...
 
       def check_if_a_service_is_running?(service)
+        return check_if_kube_is_running? if ['kubernetes', 'k8s'].include? service
         command = "systemctl status #{service}"
         with_target_vms(nil, {:single_target=>true}) do |machine|
           return machine.communicate.test(command)
         end
       end
 
+      def check_if_kube_is_running?
```
brgnepal: short name is preferred like `kube_running?`

lib/vagrant-service-manager/command.rb
```diff
@@ -102,12 +111,20 @@ def print_help(config = {})
       end
 
       def check_if_a_service_is_running?(service)
```
brgnepal: short name is preferred like `service_running?`

lib/vagrant-service-manager/command.rb
```diff
@@ -123,8 +140,10 @@ def print_all_provider_info
         end
 
         running_services.each do |e|
-          @env.ui.info("\n#{e} env:")
-          public_send("execute_#{e}_info")
+          if not ["kubernetes", "k8s"].include? e
```
brgnepal: I hope you got the use case of `KUBE_NAMES` here now :smile:  .
Also, `unless` instead of `if not`.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix#173: Shows if kubernetes services is running in the box
```
brgnepal: User name missing

lib/vagrant-service-manager/command.rb
navidshaikh, line 17: > I think 'docker' is not required 

@budhrg  : Why won't it be required?

brgnepal, line 17: @praveenkumar can answer this better as he told me during discussion.

praveenkumar, line 17: @navidshaikh @budhrg As per my understanding to make sure minion works properly we should have docker service running and in our case it already running when you do restart then docker service might not be needed because we usually need a service to be restart if service is hanged or changed some configuration. I checked config file of kubernetes-node and kubernetes-client but didn't see any docker restart requirement when those config change. Again it is debatable and as per http://kubernetes.io/docs/getting-started-guides/fedora/fedora_manual_config/ guideline they also pointed `restart docker` for minion. So I would say this is not a reason to block this PR.

brgnepal, line 17: Fair enough then.

brgnepal:
I tested the PR. It is working as expected showing `running` status of `kubernetes`.  
LGTM

lib/vagrant-service-manager/command.rb
navidshaikh, line 17: One more reason to have `docker` service status account while checking for kubernetes service is, irrespective of whether docker is always going to be running for ADB / CDK box, the module to do a certain task has to be complete and not having dependency on the assumptions. This gives us modularity while porting the module / code from one component to another, making sure that module does its job irrespective of the assumption. Basically here I want all dependency listing and their accountability for checking kube service status.

bexelbie, line 17: The above discussion seems like a great reason for having sccli be the "source of all information" about services in the box.

bexelbie:
Pending my question in #155 and possibly regardless, should we create a lib/vagrant-service-manager/services/kubernetes.rb and formalize our treatment here?

brgnepal:
@bexelbie , about `lib/vagrant-service-manager/services/kubernetes.rb`, it will be part of another PR which solves #11 . This PR only solve this issue #173 which is only status IMO.

navidshaikh:
> about lib/vagrant-service-manager/services/kubernetes.rb, it will be part of another PR which solves #11 . This PR only solve this issue #173 which is only status IMO.

That's correct.

brgnepal:
This PR is not needed if we agree on this issue https://github.com/projectatomic/vagrant-service-manager/issues/178.

navidshaikh:
> This PR is not needed if we agree on this issue #178.

This will be needed for `status` command now.

navidshaikh:
Rebased the PR on top of PRs updating the behavior of `env` and adding `status` command. Here is the updated output after rebase and solving conflicts:

```
➜  vagrant-service-manager git:(06230e0) ✗ bundle exec vagrant service-manager env

docker env:
export DOCKER_HOST=tcp://172.28.128.8:2376
export DOCKER_TLS_VERIFY=1

openshift env:
# You can access the OpenShift console on: https://172.28.128.8:8443/console
# To use OpenShift CLI, run: oc login https://172.28.128.8:8443


➜  vagrant-service-manager git:(06230e0) ✗ bundle exec vagrant service-manager env --script-readable
DOCKER_HOST=tcp://172.28.128.8:2376
DOCKER_CERT_PATH='/home/nshaikh/work/src/vagrant-service-manager/.vagrant/machines/default/virtualbox/docker'
DOCKER_TLS_VERIFY=1
DOCKER_API_VERSION=1.21
OPENSHIFT_URL=https://172.28.128.8:8443
OPENSHIFT_WEB_CONSOLE=https://172.28.128.8:8443/console


➜  vagrant-service-manager git:(06230e0) ✗ bundle exec vagrant service-manager status               
Configured services:
docker - running
openshift - running
kubernetes - stopped


➜  vagrant-service-manager git:(06230e0) ✗ vagrant ssh
Last login: Thu Apr 14 06:07:19 2016
[vagrant@rhel-cdk ~]$ sudo -i 
[root@rhel-cdk ~]# sccli k8s
Stopping the Openshift services
Removed symlink /etc/systemd/system/multi-user.target.wants/openshift.service.
Generating RSA private key, 2048 bit long modulus
...............+++
......+++
e is 65537 (0x10001)
Created symlink from /etc/systemd/system/multi-user.target.wants/etcd.service to /usr/lib/systemd/system/etcd.service.
Created symlink from /etc/systemd/system/multi-user.target.wants/kube-apiserver.service to /usr/lib/systemd/system/kube-apiserver.service.
Created symlink from /etc/systemd/system/multi-user.target.wants/kube-controller-manager.service to /usr/lib/systemd/system/kube-controller-manager.service.
Created symlink from /etc/systemd/system/multi-user.target.wants/kube-scheduler.service to /usr/lib/systemd/system/kube-scheduler.service.
Created symlink from /etc/systemd/system/multi-user.target.wants/kube-proxy.service to /usr/lib/systemd/system/kube-proxy.service.
Created symlink from /etc/systemd/system/multi-user.target.wants/kubelet.service to /usr/lib/systemd/system/kubelet.service.
[root@rhel-cdk ~]# kubectl get nodes
NAME        LABELS                             STATUS    AGE
127.0.0.1   kubernetes.io/hostname=127.0.0.1   Ready     1d
[root@rhel-cdk ~]# exit
logout
[vagrant@rhel-cdk ~]$ exit
logout
Connection to 127.0.0.1 closed.


➜  vagrant-service-manager git:(06230e0) ✗ bundle exec vagrant service-manager env
docker env:
export DOCKER_HOST=tcp://172.28.128.8:2376
export DOCKER_TLS_VERIFY=1


➜  vagrant-service-manager git:(06230e0) ✗ bundle exec vagrant service-manager status
Configured services:
docker - running
openshift - stopped
kubernetes - running


➜  vagrant-service-manager git:(06230e0) ✗ bundle exec vagrant service-manager env --script-readable
DOCKER_HOST=tcp://172.28.128.8:2376
DOCKER_CERT_PATH='/home/nshaikh/work/src/vagrant-service-manager/.vagrant/machines/default/virtualbox/docker'
DOCKER_TLS_VERIFY=1
DOCKER_API_VERSION=1.21
```

bexelbie:
LGTM

LalatenduMohanty:
May be I missed something. Does it mean if we have all these services running , the single node k8s setup is done and it is running? I guess not. That the reason I have asked the question in the beginning but I did not get the answer @navidshaikh ?

praveenkumar:
@navidshaikh @bexelbie Since I am already working on https://github.com/projectatomic/adb-utils/issues/105 so I would suggest you folks to not do it from vsm plugin. I will need 1-2 days and will send PR of a working code. Otherwise it will be duplication.

navidshaikh:
> Does it mean if we have all these services running , the single node k8s setup is done and it is running?

Yes. Do we have a case where there is an exception?

LalatenduMohanty:
@navidshaikh unless below part is not done , we will not get the single node setup.

```
sudo openssl genrsa -out /etc/pki/kube-apiserver/serviceaccount.key 2048 > /dev/null 2>&1
    sudo sed -i.back '/KUBE_API_ARGS=*/c\KUBE_API_ARGS="--service_account_key_file=/etc/pki/kube-apiserver/serviceaccount.key"' /etc/kubernetes/apiserver
    sudo sed -i.back '/KUBE_CONTROLLER_MANAGER_ARGS=*/c\KUBE_CONTROLLER_MANAGER_ARGS="--service_account_private_key_file=/etc/pki/kube-apiserver/serviceaccount.key"' /etc/kubernetes/controller-manager
```

navidshaikh:
> Since I am already working on projectatomic/adb-utils#105 so I would suggest you folks to not do it from vsm plugin

What is it that being advised not to do? is the question around verifying if kubernetes service is indeed running ?

navidshaikh:
> @navidshaikh unless below part is not done , we will not get the single node setup.

The command is more to display just the status of service, rather configuring it.
The assumption here is that, user gets the right Vagrantfile for k8s and start the service using Vagrantfile, and VSM plugin command is able to show the status of the services.

LalatenduMohanty:
@navidshaikh That means we know a case where we can fool VSM and that is a bug to me.

navidshaikh:
> That means we know a case where we can fool VSM and that is a bug to me.

There are many cases where unexpected actions in the box can cause random results. This is the sane case where we are expecting the user to go with the directed way. I am sure, in future as we'll have more progress in the plugin, we should be able to identify the cases and report them.

praveenkumar:
> What is it that being advised not to do? is the question around verifying if kubernetes service is indeed running ?

I am just giving a heads-up not stopping you to do :+1:

LalatenduMohanty:
@navidshaikh  I beg to differ with you on this. IMO we should not have gone this route. That's the reason I have asked the question in the beginning. If we know what is the end result why not checking that to see if it is setup properly.

bexelbie:
This is why I think we need to ramp up `sccli` so that it is the source of all knowledge about valid services (start, stop, restart, and verification, etc.)

praveenkumar:
@bexelbie I am on it just a preview https://github.com/projectatomic/adb-utils/issues/105 , we can put more here.

PR #186, opened 2016-04-15
Fix env lines in vagrant service-manager env

### LINKED ISSUE #184: --script-readable vs --machine-readable,  eval'able content in output
There has been a lot of regressions in the output syntax / format of service-manager env lately. There have also been new features added (ex: --script-readable) that also seem quite confusing in their implementation choices.

In an effort to clean this up to benefit clients as well as the projectatomic commiters, I'll be outlining current behavior, problems with it, and a proposal.

Definitions:

| Word | Definition |
| --- | --- |
| Plaintext | user-visible strings, not parsable, not eval'able |
| machine-readable | k=v format, xml format, anything a machine can parse without modification. |
| eval-able | something that can be eval'd and executed directly in a shell without modification |

Current state of service-manager env

1) Call to 'vagrant service-manager': 100% plaintext, usage syntax

| Command | Extra flag | Output Type |
| --- | --- | --- |
| env |  | Mixed, effectively plaintext |
| env | --machine-readable | empty |
| env | --script-readable | Broken (syntax / usage) |
| env docker |  | Kinda sorta eval-able (long command on windows) |
| env docker | --machine-readable | empty |
| env docker | --script-readable | Broken (syntax / usage) |
| env openshift |  | Plaintext |
| env openshift | --machine-readable | empty |
| env openshift | --script-readable | Properties format k=v, machine-readable |

Problem:

```
1) output format changes depending on which path you are in.
2) Not very consistent.
3) --machine-readable is the suggested flag by vagrant.
    a) You have chosen to implement --script-readable
    b) –script-readable sounds like eval'able, but for openshift it is machine-readable
    c) --script-readable for docker is broken
4) The behavior of 'service-manager env docker' 
    CANNOT BE CHANGED since JBDS/JBT now depend on it
5) 'service-manager env docker' is currently eval format, but is parsed by JBDS/JBT 
    as if it was machine-readable. Changes to this path that may work great in a shell 
    will break JBDS/JBT (ex: adding single-quotes around certificate path 8 days ago). 
6) to eval 'service-manager env docker' on windows/cygwin, you have to enter this long command:
    eval "$(VAGRANT_NO_COLOR=1 vagrant service-manager env docker | tr -d '\r')"
```

Proposal:

| Command | Extra flag | Output Type |
| --- | --- | --- |
| env |  | Mixed, effectively plaintext |
| env | --machine-readable | k=v format for docker and openshift, with user-readable strings commented |
| env | --script-readable | Setx / export syntax for docker and openshift, with user-readable strings commented |
| env docker |  | eval-able (Cannot be changed, except make more eval-able on windows? dangerous though) |
| env docker | --machine-readable | K=V properties format with user-readable strings commented |
| env docker | --script-readable | Equivalent to 'env docker' |
| env openshift |  | Plaintext |
| env openshift | --machine-readable | K=V syntax with user-readable strings commented (ie, today's –script-readable format) |
| env openshift | --script-readable | Setx / export syntax for openshift, with user-readable strings commented |

Finally, it should be noted that all changes to 'env docker' (with no extra flags) are now 'dangerous'.  While no flags such as --machine-readable or --script-readable are being passed in, the JBDS/JBT current toolset is now dependent on this exact output format, and somewhat poorly, as well.  Even small changes like adding a single-quote to parts of the output could break our parsing, since we can't know what types of changes are being made there.

bexelbie:
They are now evalable comments
Fix #184

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -179,7 +179,7 @@ module Vagrant
         running_services.each do |e|
           unless  script_readable
             # since we do not have feature to show the kube connection information
-            @env.ui.info("\n#{e} env:") unless KUBE_NAMES.include? e
+            @env.ui.info("\n# #{e} env:") unless KUBE_NAMES.include? e
           end
           public_send("execute_#{e}_info", script_readable) unless KUBE_NAMES.include? e
         end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix#173: Shows if kubernetes services is running in the box @navidshaikh
 - Fix #169: Adds command for displaying box routable IP address @navidshaikh
 - Fix message for box command on default help @budhrg
+- Fix #184: Make env headers comments for vagrant service-manager env @bexelbie
 
 ## v1.0.1 Apr 12, 2016
 - Updated SPEC (v1.0.0) for url, date and format @budhrg


bexelbie:
Sample Output:

```
---> vagrant service-manager env

# docker env:
export DOCKER_HOST=tcp://192.168.121.190:2376
export DOCKER_CERT_PATH=/home/bexelbie/Repositories/projectatomic/vagrant-service-manager/.vagrant/machines/default/libvirt/docker
export DOCKER_TLS_VERIFY=1

# openshift env:
# You can access the OpenShift console on: https://192.168.121.190:8443/console
# To use OpenShift CLI, run: oc login https://192.168.121.190:8443
```

lib/vagrant-service-manager/command.rb
```diff
@@ -179,7 +179,7 @@ def print_all_provider_info(script_readable = false)
         running_services.each do |e|
           unless  script_readable
             # since we do not have feature to show the kube connection information
-            @env.ui.info("\n#{e} env:") unless KUBE_NAMES.include? e
+            @env.ui.info("\n##{e} env:") unless KUBE_NAMES.include? e
```
brgnepal: Need a space between two `#`.

brgnepal:
@bexelbie Could you pull my PR, change it and send PR against my branch budhrg/fix-135 ?

hferentschik:
I still argue that the whole `service-manager env` business is wrong and trying to list the result of all sub env calls is bad design out of a misguided "performance" argument. It's two less than two seconds you save (initializing the Ruby runtime). But anyways, ...

I was wondering whether the output of the `env` and `status` calls are not a perfect candidate of adding acceptance tests. Put once and for all the requirements into some tangible form. WDYT?

bexelbie:
@hferentschik 

> I still argue that the whole service-manager env business is wrong and trying to list the result of all sub env calls is bad design out of a misguided "performance" argument. It's two less than two seconds you save (initializing the Ruby runtime). But anyways, ...

I don't follow you here.  I believe I am hearing the following statements all at once and I don't understand how they all fit together:
1. `vagrant service-manager env` should output all environment information for all running services
2. `vagrant service-manager env foo` should not be supported
3. `vagrant service-manager env` is misguided and poorly designed

Can you help me understand exactly what you are saying?

hferentschik:
> Can you help me understand exactly what you are saying?

Sure. The suggested patch will make the output of the command evalable. AFAIU that's what the majority thinks is the right thing to do. So lets merge this.

---

For the records:

>  vagrant service-manager env should output all environment information for all running services

That's not my opinion. I actually disagree.

> vagrant service-manager env foo should not be supported

No, no. That's a misunderstanding. On the contrary, I would only support the explicit calls

> vagrant service-manager env is misguided and poorly designed

Right.

navidshaikh:
> That's not my opinion. I actually disagree.

@hferentschik 
[Here](https://github.com/projectatomic/vagrant-service-manager/issues/184#issuecomment-209970228) is the proposal from @robstryker stating that

```
env     eval'able info from docker and openshift
```

I filed https://github.com/projectatomic/vagrant-service-manager/issues/194 to finalize on the behavior of `vagrant service-manager env`. 

I will merge this PR and I think the resolution of https://github.com/projectatomic/vagrant-service-manager/issues/194 should fix the concern.

PR #190, opened 2016-04-18
Fix #188: Name of k8s service not consistent

### LINKED ISSUE #188: name of k8s service not consistent
`vagrant service-manager env` prints "kubernetes" in 1st section,
but for `vagrant service-manager restart`, user need to pass "k8s" argument for restarting kubernetes.
This can be quite misleading, I would like to have the same keyword for both scenarios.

brgnepal:
Fix #188 
- Removed k8s now as we will only support 'kubernetes'

It should be merged once this issue https://github.com/projectatomic/adb-utils/issues/102 is fixed.

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -6,8 +6,7 @@ module Vagrant
   module ServiceManager
     DOCKER_PATH = '/home/vagrant/.docker'
     SUPPORTED_SERVICES = ['docker', 'openshift', 'kubernetes']
-    SCCLI_SERVICES = ['openshift', 'k8s']
-    KUBE_NAMES = ['kubernetes', 'k8s']
+    SCCLI_SERVICES = ['openshift', 'kubernetes']
     KUBE_SERVICES = [
       'etcd', 'kube-apiserver', 'kube-controller-manager', 'kube-scheduler',
       'kubelet', 'kube-proxy', 'docker'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Updated README to make Installation Instructions clearer @bexelbie
 - Fix #195 Adding Cucumber and Aruba based acceptance tests @hferentschik
 - CHANGELOG fix and README update for OS support for tests @budhrg
+- Fix #188: Name of k8s service not consistent @budhrg
 
 ## v1.0.2 May 09, 2016
 - Add --script-readable to env and env docker @bexelbie


navidshaikh:
I remember we adding support for accepting either `kubernetes` or `k8s` as aliases for kubernetes provider and support for restarting kubernetes with either of names however `sccli` only understands `k8s`. @LalatenduMohanty and I had a discussion regarding this, and he suggested we follow what `sccli` provides.

I'd still like to support either of names for kube service and since this came up again, ping @LalatenduMohanty

bexelbie:
Why are we supporting the k8s name at all?  What was the requirement on that?  It seems odd to have that be preferred over the proper name of the service.

If we want to support k8s, I believe it should be dealiased to kubernetes and all tools should actually operate on that term.

brgnepal:
@praveenkumar, @LalatenduMohanty you thoughts?

praveenkumar:
@budhrg we are going to remove `k8s` from sccli, it should be `kubernetes` as @bexelbie said we shouldn't prefer odd name over a proper name for a service.

brgnepal:
@praveenkumar : Then, I will update this PR for `kubernetes`  term and wait till we get its respective fix in ADB box. Please ACK.

navidshaikh:
> Then, I will update this PR for kubernetes term and wait till we get its respective fix in ADB box. Please ACK.

ACK!

navidshaikh:
Here is the respective issue for adb-utils https://github.com/projectatomic/adb-utils/issues/102

lib/vagrant-service-manager/command.rb
```diff
@@ -6,8 +6,7 @@ module Vagrant
   module ServiceManager
     DOCKER_PATH = '/home/vagrant/.docker'
     SUPPORTED_SERVICES = ['docker', 'openshift', 'kubernetes']
-    SCCLI_SERVICES = ['openshift', 'k8s']
-    KUBE_NAMES = ['kubernetes', 'k8s']
+    SCCLI_SERVICES = SUPPORTED_SERVICES.drop(1)
```
navidshaikh: Can we define the `SCCLI_SERVICES` explicitly instead of operating on `SUPPORTED_SERVICES`
like

```
SUPPORTED_SERVICES = ["openshift", "kubernetes"]
```

navidshaikh:
The `restart` command should work for `kubernetes`, this requires that `sccli` understands `kubernetes` as the name of Kubernetes  (earlier adb-utils used to operate on name `k8s`, and now `kubernetes` see https://github.com/projectatomic/adb-utils/issues/102#issuecomment-215051316)

brgnepal:
This PR requires ADB 2.1 and adb-utils 1.6 versions. (Based on [this comment](https://github.com/projectatomic/adb-utils/issues/102#issuecomment-217824494))

brgnepal:
@navidshaikh @bexelbie : Rebased with upstream and ready to merge as ADB 2.1 and adb-utils 1.6 are released.

Issue #195, opened 2016-04-21
Create basic acceptance tests for service-manager


hferentschik:

hferentschik:
Relates to issue #194

optak:
not sure what you mean by acceptance test, but we have sanity test for service-manager env here:
https://github.com/projectatomic/adb-tests/blob/master/testsuite/Components/vagrant/service-manager/sanity/runtest.sh

hferentschik:
> not sure what you mean by acceptance test, but we have sanity test for service-manager env here

Right, something along these lines, but using vagrant-spec, aka a Ruby testing framework. The plugin is written in Ruby and it makes sense to stay in this idiom. Also one has all the universe of Ruby to write tests. Also the tests would be living with the code, so the developer can use them during development (actually being able to debug the code), but also to verify that all tests are still passing when creating a pull request. In a perfect world, we then even have a CI server which picks up the pull request and verifies it against the test. @optak, on that note, do we have infrastructure for a CI server? 

BTW, it will look very similar to this - https://github.com/projectatomic/adb-vagrant-registration/pull/70/commits/81ed5c61aed710194560c2192cdc2334c682e65a

The idea is to be able to run the tests out of the box against VirtualBox and Libvirt.

optak:
Thanks for info and example @hferentschik. So you are talking about unit tests? In CDK QE we focus more on integration and testing on different environment, so it could be:
1) unit tests run upon git commit and written in the same language as tested code
2) end2end scenarios, integration testing: use whatever framework and language (bash + beakerlib is not final, it can change), living in https://github.com/projectatomic/adb-tests, in future with some CI infrastructure as well.
Running all tests from that repo takes probably too much time for testing each commit anyway.

hferentschik:
> So you are talking about unit tests? 

Unit tests as well, yes. They would most likely use be using straight RSpec tests. vagrant-specs are really acceptance tests. Or call it black box tests if you like. They make vagrant plugin calls expecting certain outcome from it w/o knowing how or what the plugin does. Are these acceptance tests as per QA. I guess they could and IMO they could be reused, but that's a different discussion.

> In CDK QE we focus more on integration

the vagrant-spec tests will be integration tests

> and testing on different environment

That part is imo most critical. That's not something covered directly by these tests. Obviously these test should run on various environment (the one the dev works on), but they are not trying to somehow automatically run on various platform (eg via virtualization or so).

[...]

PR #202, opened 2016-04-26
Issue 195 Acceptance test for service-manager using vagrant-spec

hferentschik:
Here is a first cut of the acceptance test harness. I want to add some more tests and maybe make some other adjustments, but I also want to share it early to get some feedback and making sure that everyone can run it.

The tests can be run against VirtualBox or Libvirt (the latter with a caveat). If you have the box file you want to use for testing already on your system, you can run the tests like:

```
$ export VAGRANT_SPEC_BOX=file://<full path to box file>
# VirtualBox
$ bundle exec rake acceptance
# Libvirt
$ bundle exec rake acceptance['libvirt']
```

Obviously the box file used needs to be the correct one for the provider. I added another helper task which can download the publicly available box files (might be nice for CI testing!?)

```
$ export REDHAT_USER=<any user who can login into access.redhat.com>
$ export REDHAT_PASSWORD=<mypass>
$ bundle exec rake getbox
# or
$ bundle exec rake getbox['libvirt']
```

Regarding the mentioned caveat. The current tests pass on OS X using the VirtualBox box. When I run the same test on Linux using Libvirt, I get one test failure. The gist seems to be that when using VirtualBox I get _DOCKER_API_VERSION=1.21_ as part of the docker env output whereas with Libvirt I get _DOCKER_API_VERSION=1.20_. 

WDYT?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,13 +3,20 @@ source 'https://rubygems.org'
 gemspec
 
 group :development do
-  gem 'vagrant', git: 'https://github.com/mitchellh/vagrant.git'
+  gem 'vagrant',      git: 'https://github.com/mitchellh/vagrant.git'
+  gem 'vagrant-spec', git: 'https://github.com/mitchellh/vagrant-spec.git'
+  gem 'gem-compare'
+  gem 'rake'
+  gem 'vagrant-vbguest'
+  gem 'vagrant-libvirt'
+  gem 'fog-libvirt', '0.0.3' # https://github.com/pradels/vagrant-libvirt/issues/568
+  gem 'mechanize'
+
   # added as the vagrant component wouldn't build without it
   gem 'json'
-  gem 'rake'
-  gem 'bundler', '~> 1.6'
 end
 
 group :plugins do
   gem 'vagrant-service-manager', path: '.'
+  gem 'vagrant-registration'
 end
--- a/Rakefile
+++ b/Rakefile
@@ -1,1 +1,63 @@
-require "bundler/gem_tasks"
+require 'bundler/gem_tasks'
+require 'rake/clean'
+require 'mechanize'
+require 'fileutils'
+
+CLOBBER.include('pkg')
+CLEAN.include('build')
+
+desc 'Download Vagrant box'
+task :getbox, [:provider] do |t, args|
+  if args[:provider].nil?
+    provider='virtualbox'
+  else
+    provider=args[:provider]
+  end
+  agent = Mechanize.new
+  agent.follow_meta_refresh = true
+  agent.get('https://access.redhat.com/downloads/content/293/ver=2/rhel---7/2.0.0/x86_64/product-software') do |page|
+
+    # Submit first form which is the redirect to login page form
+    login_page = page.forms.first.submit
+
+    # Submit the login form
+    after_login = login_page.form_with(:name => 'login_form') do |f|
+     username_field = f.field_with(:id => 'username')
+     username_field.value = ENV['REDHAT_USER']
+     password_field = f.field_with(:id => 'password')
+     password_field.value = ENV['REDHAT_PASSWORD']
+    end.click_button
+
+    # There is one more redirect afte successful login
+    download_page = after_login.forms.first.submit
+
+    download_page.links.each do |link|
+     if link.href =~ /rhel-cdk-kubernetes-7.2-23.x86_64.vagrant-#{Regexp.quote(provider)}.box/
+      puts "Downloading box #{link.href}"
+      download_dir = File.join(File.dirname(__FILE__), 'build', 'tmp')
+      unless File.directory?(download_dir)
+       FileUtils.mkdir_p(download_dir)
+      end
+      agent.pluggable_parser.default = Mechanize::Download
+      agent.get(link.href).save(File.join(download_dir, 'cdk.box'))
+     end
+    end
+  end
+end
+
+desc 'Run acceptance specs using vagrant-spec'
+task :acceptance, [:provider] do |t, args|
+  if args[:provider].nil?
+    provider='virtualbox'
+  else
+    provider=args[:provider]
+  end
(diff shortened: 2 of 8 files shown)


brgnepal:
@hferentschik Could you squash into one commit?

hferentschik:
> Could you squash into one commit?

Why? There is a reason for the current commit structure. I am happy to move the Gemfile.lock into the main commit, provided we agree that this is the right thing to do. 

The separate README cleanup commit should stay separate though imo.

brgnepal:
@hferentschik It would be easy to track in future by one commit what all files were changed rather than looking for individual ones. Its like name-spacing all your changes related to issue fix under PR subject commit message "Fix #195 Acceptance test for service-manager using vagrant-spec".

May be @LalatenduMohanty could add more lights on it.

hferentschik:
> It would be easy to track in future by one commit what all files were changed rather than looking for individual ones.

True. There is a middle ground, but usually I try to separate "cleanup" and refactoring commits from the commits which change functionality. Squashing all commits for a given issue into a single commit is rarely a good idea. It also makes reviewing harder. That said, there is somewhere a middle ground. Micro commits are useless as well. 

Also, as I mentioned, this is still work in progress and while it is I am not going to squash anything. Squashing is easy, splitting again out into multiple commits is much harder. So until we agree on the final details of the change, I'll keep the separate commits. Whether in the end it will get applied as single commit or multiple we can decide then

>  Its like name-spacing all your changes related to issue fix under PR subject commit message

This is exactly what I am doing. All my commits related to this issue have the "Issue #<key>" prefix.
In the Hibernate projects we had a policy to separate typo fixes, refactoring and so on from the actual issue fix. And even the issue fix itself could be multiple commits depending on the size of the change. Important is just that each commit still is "runnable" and passes the tests (think git bisect).

BTW, instead of whether to squash or not to squash we should talk about merge commits vs linear history. You want to improve your git history, I think we should adopt a linear, fast forward history only.

Rakefile
```diff
@@ -1 +1,63 @@
...
+desc 'Download Vagrant box'
+task :getbox, [:provider] do |t, args|
+  if args[:provider].nil?
+    provider='virtualbox'
+  else
+    provider=args[:provider]
+  end
+  agent = Mechanize.new
+  agent.follow_meta_refresh = true
+  agent.get('https://access.redhat.com/downloads/content/293/ver=2/rhel---7/2.0.0/x86_64/product-software') do |page|
```
brgnepal, line 18: Would be better to extract it out in some constant like `RHEL_DOWNLOAD_LINK`

hferentschik:
> Would be better to extract it out in some constant like `RHEL_DOWNLOAD_LINK`

+1 good idea

LalatenduMohanty:
@hferentschik I think the  Gemfile.lock patch should be an independent patch. If you send the patch separately it is most likely to get merged relatively quicker than this patch too.

LalatenduMohanty:
@hferentschik Have you given the patch a spin with ADB Vagrant box?

Rakefile
LalatenduMohanty, line 18: And also adding a constant for  ADB download link . Not sure if we can use atlas. But we can use http://cloud.centos.org/centos/7/atomic/images/AtomicDeveloperBundle-1.8.0-CentOS7-VirtualBox.box

Gemfile
```diff
@@ -3,13 +3,20 @@ source 'https://rubygems.org'
 gemspec
 
 group :development do
-  gem 'vagrant', git: 'https://github.com/mitchellh/vagrant.git'
+  gem 'vagrant',      git: 'https://github.com/mitchellh/vagrant.git'
+  gem 'vagrant-spec', git: 'https://github.com/mitchellh/vagrant-spec.git'
+  gem 'gem-compare'
+  gem 'rake'
+  gem 'vagrant-vbguest'
```
LalatenduMohanty, line 10: We don't need vagrant-vbguest here.

Gemfile
```diff
@@ -3,13 +3,20 @@ source 'https://rubygems.org'
...
+  gem 'vagrant-libvirt'
```
LalatenduMohanty, line 11: This is also not compulsory for Virtualbox provider

Gemfile
```diff
@@ -3,13 +3,20 @@ source 'https://rubygems.org'
...
+
   # added as the vagrant component wouldn't build without it
   gem 'json'
-  gem 'rake'
-  gem 'bundler', '~> 1.6'
 end
 
 group :plugins do
   gem 'vagrant-service-manager', path: '.'
+  gem 'vagrant-registration'
```
LalatenduMohanty, line 21: vagrant-registration is not required for ADB.

Rakefile
hferentschik, line 18: > And also adding a constant for ADB download link 

+1 I forgot about ADB. I'll have to tasks. One for CDK and one for ADB.

hferentschik:
> I think the Gemfile.lock patch should be an independent patch. If you send the patch separately it is most likely to get merged relatively quicker than this patch too.

+1 I think it is a good idea to track this separately in a different issue.

Gemfile
hferentschik, line 10: true true,  cut&paste error from vagrant-registration. I'll remove it.

Gemfile
hferentschik, line 11: > This is also not compulsory for Virtualbox provider

what is not? Note, these gems are in the development group, so there is no harm for the actual runtime dependencies.

Gemfile
hferentschik, line 21: That's an interesting one, for the CDK one needs it, even though one might get away without registration altogether for the things we want to test. I can give it a go.

hferentschik:
> Have you given the patch a spin with ADB Vagrant box?

Not yet. I need to see what's the best way to do that. Either two separate specs and potentially Vagrant skeletons and share some common test file or a single spec file as we have now.

hferentschik:
Closing this for now. Trying something else as well...

PR #218, opened 2016-05-17
Add "start" as an alias for "restart"

bexelbie:
This allows some instructions and concepts to "read" better.

Helps with https://github.com/projectatomic/adb-atomic-developer-bundle/pull/386

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -102,7 +102,7 @@ module Vagrant
           else
             print_help(type: command, exit_status: 1)
           end
-        when 'restart'
+        when 'restart', 'start'
           exit_if_machine_not_running
           case subcommand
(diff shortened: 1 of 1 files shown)


bexelbie:
@budhrg

LalatenduMohanty:
IMO restart and start are two different things, so we should not generalize both.

bexelbie:
@LalatenduMohanty can you provide some guidance on how you see the code that implements them differing?  I didn't see any reason to modify the code.

brgnepal:
@bexelbie , @LalatenduMohanty : With this PR https://github.com/projectatomic/vagrant-service-manager/pull/219 , we may not need this PR since setting alias is confusing and under the hood it is performing `restart` operation.

bexelbie:
Closing to be replaced by #219

PR #219, opened 2016-05-17
Fix #172: Implement "start/enable" service command

### LINKED ISSUE #172: Implement "start/enable" service command
Hi,

We need to add a way to start or enable service as:

```
$ vagrant service-manager start <service>
```

to start or enable service.

brgnepal:
Fix #172.
- Mapped `start/stop/restart/status` commands with
  `sccli`(https://github.com/projectatomic/adb-utils/blob/master/utils/sccli.py)
- Added `stop` operation
- Removed support of `systemd` services. Only sccli services are
  supported
- Updated implementation of `status` as per `sccli`
- Added acceptance tests
- Updated message if service name is missing
- Updated help messages

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,7 +10,7 @@ CDK_DOWNLOAD_URL='https://access.redhat.com/downloads/content/293/ver=2/rhel---7
 CDK_BOX_BASE_NAME='rhel-cdk-kubernetes-7.2-23.x86_64.vagrant'
 
 ADB_DOWNLOAD_URL='http://cloud.centos.org/centos/7/atomic/images'
-ADB_BOX_BASE_NAME='AtomicDeveloperBundle-2.0.0-CentOS7'
+ADB_BOX_BASE_NAME='AtomicDeveloperBundle-2.1.0-CentOS7'
 
 CLOBBER.include('pkg')
 CLEAN.include('build')
--- a/features/service-operation.feature
+++ b/features/service-operation.feature
@@ -1,0 +1,40 @@
+Feature: Command output from service operations like stop/start/restart
+  service-manager should return the correct exit code from stop/start/restart command
+
+  @operation
+  Scenario Outline: Boot and execute service operations like stop/start/restart
+    Given box is <box>
+    And provider is <provider>
+    And a file named "Vagrantfile" with:
+    """
+    require 'vagrant-libvirt'
+
+    Vagrant.configure('2') do |config|
+      config.vm.box = '<box>'
+      config.vm.box_url = 'file://../boxes/<box>-<provider>.box'
+      config.vm.network :private_network, ip: '<ip>'
+      config.vm.synced_folder '.', '/vagrant', disabled: true
+      config.servicemanager.services = 'docker'
+    end
+    """
+
+    When I successfully run `bundle exec vagrant up --provider <provider>`
+      When the "docker" service is running
+      When I successfully run `bundle exec vagrant service-manager stop docker`
+      Then the service "docker" should be stopped
+
+      When the "docker" service is not running
+      And I successfully run `bundle exec vagrant service-manager start docker`
+      Then the service "docker" should be running
+
+      When the "docker" service is running
+      And I successfully run `bundle exec vagrant service-manager restart docker`
+      Then the service "docker" should be running
+      And have a new pid for "docker" service
+
+    Examples:
+      | box   | provider   | ip          |
+      | cdk   | virtualbox | 10.10.10.42 |
+      | adb   | virtualbox | 10.10.10.42 |
+      | cdk   | libvirt    | 10.10.10.42 |
+      | adb   | libvirt    | 10.10.10.42 |
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -49,6 +49,11 @@ def output_is_script_readable(raw_stdout)
   end
 end
 
+def extract_process_id(data)
+  tokens = data.scan(/Main PID: ([0-9]+) \(/)
+  tokens.last.first.to_i unless tokens.empty?
+end
+
 ###############################################################################
 # Some shared step definitions
 ##############################################################################
@@ -80,4 +85,54 @@ end
 
 Then(/^stdout from "([^"]*)" should match \/(.*)\/$/) do |cmd, regexp|
   aruba.command_monitor.find(Aruba.platform.detect_ruby(cmd)).send(:stdout) =~ /#{regexp}/
+end
+
+# track service process ID
+@service_current_process_id = -1
+
+# Note: Only for services supported through systemctl. Not for 'kubernetes'.
+Then(/^the service "([^"]*)" should be ([^"]*)$/) do |service, operation|
+  run("vagrant ssh -c \"sudo systemctl status #{service}\"")
+
+  if ['running', 'restarted'].include? operation
+    exit_code = 0
+    regexp = /Active: active \(running\)/
+  elsif operation == 'stopped'
+    exit_code = 3
+    regexp = /Active: inactive\(dead\)/
+  end
+
+  expect(last_command_started).to have_exit_status(exit_code)
+  aruba.command_monitor.find(Aruba.platform.detect_ruby(last_command_started)).send(:stdout) =~ regexp
+end
+
+# Note: Only for services supported through systemctl. Not for 'kubernetes'.
+When(/^the "([^"]*)" service is( not)? running$/) do |service, negated|
+  run("vagrant ssh -c \"sudo systemctl status #{service}\"")
+
+  if negated
+    expect(last_command_started).to have_exit_status(3)
+  else
+    expect(last_command_started).to have_exit_status(0)
+    stdout = aruba.command_monitor.find(Aruba.platform.detect_ruby(last_command_started)).send(:stdout)
+    @service_current_process_id = extract_process_id(stdout)
+  end
+end
+
+# Note: Only for services supported through systemctl. Not for 'kubernetes'.
+When(/^the "([^"]*)" service is \*not\* running$/) do |service|
+  # Stop the service
+  run("vagrant ssh -c \"sudo systemctl stop #{service}\"")
+
+  expect(last_command_started).to have_exit_status(0)
+  step "the \"docker\" service is not running"
+end
+
+# Note: Only for services supported through systemctl. Not for 'kubernetes'.
+Then(/^have a new pid for "([^"]*)" service$/) do |service|
+  run("vagrant ssh -c \"sudo systemctl status #{service}\"")
+
+  expect(last_command_started).to have_exit_status(0)
+  stdout = aruba.command_monitor.find(Aruba.platform.detect_ruby(last_command_started)).send(:stdout)
+  expect(@service_current_process_id).not_to eq(extract_process_id(stdout))
 end
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -7,7 +7,6 @@ module VagrantPlugins
   module ServiceManager
     DOCKER_PATH = '/home/vagrant/.docker'
     SUPPORTED_SERVICES = ['docker', 'openshift', 'kubernetes']
-    SCCLI_SERVICES = ['openshift', 'kubernetes']
     KUBE_SERVICES = [
       'etcd', 'kube-apiserver', 'kube-controller-manager', 'kube-scheduler',
       'kubelet', 'kube-proxy', 'docker'
@@ -111,13 +110,13 @@ module VagrantPlugins
           else
             print_help(type: command, exit_status: 1)
           end
-        when 'restart'
+        when 'restart', 'start', 'stop'
           exit_if_machine_not_running
           case subcommand
           when '--help', '-h'
-            print_help(type: command)
+            print_help(type: 'operation', operation: command)
           else
-            restart_service(subcommand)
+            perform_service(command, subcommand)
           end
         when '--help', '-h', 'help'
           print_help
@@ -136,7 +135,7 @@ module VagrantPlugins
         config[:type] ||= 'default'
         config[:exit_status] ||= 0
 
-        @env.ui.info(I18n.t("servicemanager.commands.help.#{config[:type]}"))
+        @env.ui.info(I18n.t("servicemanager.commands.help.#{config[:type]}", operation: config[:operation]))
         exit config[:exit_status]
       end
 
@@ -181,20 +180,21 @@ module VagrantPlugins
         end
       end
 
-      def restart_service(service)
+      def perform_service(operation, service)
         if service.nil?
-          help_msg = I18n.t('servicemanager.commands.help.restart')
-          service_missing_msg = I18n.t('servicemanager.commands.restart.service_missing')
-          @env.ui.error help_msg.gsub(/Restarts the service/, service_missing_msg)
+          help_msg = I18n.t('servicemanager.commands.help.operation', operation: operation)
+          service_missing_msg = I18n.t('servicemanager.commands.operation.service_missing')
+          @env.ui.error help_msg.gsub(/#{operation}s the service/, service_missing_msg)
           exit 126
         end
 
-        command = if SCCLI_SERVICES.include? service
+        command = if SUPPORTED_SERVICES.include? service
                     # TODO : Handle the case where user wants to pass extra arguments
                     # to OpenShift service
-                    "sccli #{service}"
+                    "sccli #{service} #{operation}"
                   else
-                    "systemctl restart #{service}"
+                    @env.ui.error I18n.t('servicemanager.commands.operation.sccli_only_support')
+                    exit 126
                   end
 
         with_target_vms(nil, single_target: true) do |machine|
--- a/locales/en.yml
+++ b/locales/en.yml
@@ -24,7 +24,9 @@ en:
           Commands:
                env          displays connection information for services in the box
                box          displays box related information like version, release, IP etc
-               restart      restarts the given systemd service in the box
+               restart      restarts the given service in the box
+               start        starts the given service in the box
+               stop         stops the given service in the box
                status       list services and their running state
 
           Options:
@@ -69,20 +71,20 @@ en:
 
           Example:
                 vagrant service-manager status openshift
-        restart: |-
-          Restarts the service
+        operation: |-
+          %{operation}s the service
 
-          Usage: vagrant service-manager restart <service> [options]
+          Usage: vagrant service-manager %{operation} <service> [options]
 
           Service:
-              A service provided by sccli or systemd. For example:
-               docker, k8s, or openshift etc.
+              A service provided by sccli. For example:
+               docker, kubernetes, openshift etc
 
           Options:
                 -h, --help         print this help
 
           Examples:
-            vagrant service-manager restart docker
+            vagrant service-manager %{operation} docker
 
       env:
         docker:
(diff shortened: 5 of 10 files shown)


lib/vagrant-service-manager/command.rb
```diff
@@ -183,7 +183,7 @@ def restart_service(service)
                     # to OpenShift service
                     "sccli #{service}"
```
bexelbie: This could be a problem.  People may expect each operation to do somethign specific and with an sccli service they do not.   This goes to what @hferentschik was saying about needing a way to separate "generic services" from "core services" (my words).

locales/en.yml
```diff
@@ -75,7 +75,9 @@ en:
 
           Service:
               A service provided by sccli or systemd. For example:
-               docker, k8s, or openshift etc.
+               sccli service like docker, k8s, openshift etc
```
bexelbie: We need to better scope this help as I think `sccli` service is not an easily understandable term.  I'll think on it

bexelbie:
Please update this PR to include rewriting the cockpit instructions to reflect the change.

lib/vagrant-service-manager/command.rb
brgnepal: Agree on it. 
@praveenkumar : Any possibility to have such feature in `sccli` in future?

brgnepal:
> Please update this PR to include rewriting the cockpit instructions to reflect the change.

Didn't get where to add cockpit instructions? Are you talking about this instructions (https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/cockpit.rst)

bexelbie:
@budhrg Yep

bexelbie:
@budhrg So I just realized this is in another repo. Was thinking about the code as a whole. Doh!

locales/en.yml
brgnepal: @bexelbie Ya agreed.

praveenkumar: Currently `sccli` able to understand docker, kubernetes and openshift. It would be nice if you folks can try latest version of adb-utils which have improved sccli.

lib/vagrant-service-manager/command.rb
praveenkumar: > Separate "generic services" from "core services"

So my understanding is currently sccli support core services (docker, openshift and kubernetes) and for generic we can directly use systemd. We are using sccli for services which require some configuration before starting it.

bexelbie: Agreed.  This needs to be implemented in vagrant-service-manager, not adb-utils.

brgnepal:
Updated to reflect the [new `sccli`](https://github.com/projectatomic/adb-utils/blob/master/utils/sccli.py) changes.

My trial for [Kubernetes Vagrantfile](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/components/centos/centos-k8s-singlenode-setup/Vagrantfile) here:

```
➜  vagrant-service-manager git:(fix-172) ✗ bundle exec vagrant service-manager status
Configured services:
docker - running
openshift - stopped
kubernetes - running

➜  vagrant-service-manager git:(fix-172) ✗ bundle exec vagrant service-manager stop kubernetes

➜  vagrant-service-manager git:(fix-172) ✗ bundle exec vagrant service-manager status kubernetes
kubernetes - stopped

➜  vagrant-service-manager git:(fix-172) ✗ bundle exec vagrant service-manager start kubernetes

➜  vagrant-service-manager git:(fix-172) ✗ bundle exec vagrant service-manager status kubernetes
kubernetes - running

```

locales/en.yml
navidshaikh: shouldn't the kubernetes service name be `kubernetes` instead of `k8s`?

```
A service provided by sccli or systemd. For example:
sccli service like docker, kubernetes, openshift etc
```

navidshaikh:
apart from https://github.com/projectatomic/vagrant-service-manager/pull/219#discussion_r64058694 rest LGTM

lib/vagrant-service-manager/command.rb
```diff
@@ -180,9 +180,9 @@ def restart_service(service)
         command = if SCCLI_SERVICES.include? service
                     # TODO : Handle the case where user wants to pass extra arguments
                     # to OpenShift service
-                    "sccli #{service}"
+                    "sccli #{service} #{operation}"
                   else
-                    "systemctl restart #{service}"
+                    "systemctl #{operation} #{service}"
```
bexelbie: Do we need to allow for enable and disable?

navidshaikh: @praveenkumar : Any recommendation on handling for `enable` and `disable` operation of services via `sccli` ?

brgnepal: As per @praveenkumar comment on `enable` and `disable`, ADB mostly provided services in the box and these operations may be defined on them.

brgnepal:
Updated with suggested changes along with acceptance test. 

@hferentschik Please review.

README.md
```diff
@@ -104,9 +104,13 @@ The following section lists the available commands for the plugin and their expl
 
    Lists services and their running state. If a `service` is specified only the status of that service is displayed. If no service is provided then only supported orchestrators are reported.
 
-6. `vagrant service-manager restart [service]`
+6. `vagrant service-manager [operation] [service]`
```
hferentschik: So what's service here? docker, openshift, kubernets, (mesos)?

locales/en.yml
```diff
@@ -75,7 +75,9 @@ en:
 
           Service:
               A service provided by sccli or systemd. For example:
-               docker, k8s, or openshift etc.
+               sccli service like docker, kubernetes, openshift etc
+               or
+               systemd service like libvirtd etc
```
hferentschik: so we are going ahead to control any systemd service now? Has this been discussed? This is imo not consistent with for example `vagrant service-manager status`. -1 from me. Was the plan not to discuss this first on the mailing list?

hferentschik:
I don't seem to be able to see an update to the global help. That would need updating as well, right?

features/service-operation.feature
```diff
@@ -0,0 +1,36 @@
...
+      config.vm.box = '<box>'
+      config.vm.box_url = 'file://../boxes/<box>-<provider>.box'
+      config.vm.network :private_network, ip: '<ip>'
+      config.vm.synced_folder '.', '/vagrant', disabled: true
+      config.servicemanager.services = 'docker'
+    end
+    """
+
+    When I successfully run `bundle exec vagrant up --provider <provider>`
+      And I successfully run `bundle exec vagrant service-manager stop docker`
```
hferentschik: These tests are "verifying" the existence of the start/stop/restart command which is good, but they don't check at all, if they work at all. We should actually verify the state on the VM as well and check that the commands do what we want. 

You can run commands on the VM like this:

```
run("vagrant ssh -c \"systemd status docker\"")
expect(last_command_started).to have_exit_status(0)
expect(last_command_started).to have_output(/<whatver makes sense to match in order to verify that the service is started/stopped/restarted>/)
```

See also https://github.com/projectatomic/adb-vagrant-registration/pull/70/files

hferentschik: You could also pack the service into the outline and make this test run against the various services, at least Docker and OpenShift!?

locales/en.yml
brgnepal: @hferentschik It was been discuss when we are implementing `restart` feature. If you see the [code implementation](https://github.com/projectatomic/vagrant-service-manager/blob/master/lib/vagrant-service-manager/command.rb#L185), we are supporting `systemd` service through CLI. Its kind of added advantage using our plugin.

cc @bexelbie

features/service-operation.feature
brgnepal: Openshift will take lot of time and will not be feasible IMO. Hence, running it against `docker`.

README.md
brgnepal: Done @hferentschik

features/service-operation.feature
hferentschik: Sure, just running against docker is fine. But you still should verify that the bahavior is correct.
Regarding OpenShift, maybe only test start/stop/restart against the CDK, there OpenShift is enabled per default. On ADB it OpenShift testing is indeed a bit of a pita, since it needs to pull the image. We should find a way to do it though. One could write the test, but only run it on demand, using an annotation.

brgnepal: Even we cannot test `restart` command due to [this issue](https://github.com/projectatomic/adb-atomic-developer-bundle/issues/342). However, to test this command, probably one of provider could be used instead of checking all.

hferentschik: The referenced issue will go away shortly. Once we update to OSE 3.2 we won't have the problems of re-occurring image pulls anymore.

hferentschik: systemctl with docker:

```
systemctl status docker
● docker.service - Docker Application Container Engine
   Loaded: loaded (/usr/lib/systemd/system/docker.service; enabled; vendor     preset: disabled)
  Drop-In: /usr/lib/systemd/system/docker.service.d
           └─flannel.conf
   Active: active (running) since Fri 2016-05-20 06:29:18 EDT; 2 days ago
     Docs: http://docs.docker.com
 Main PID: 12949 (sh)
   Memory: 115.0M
   CGroup: /system.slice/docker.service
           ├─12949 /bin/sh -c /usr/bin/docker daemon $OPTIONS                $DOCKER_STORAGE_OPTIONS            $DOCKER_NETWORK_OPTIONS                $ADD_REGISTRY            $BLOCK_REGISTRY                $INSECURE_REGISTRY            2>&1 | /usr/bin/forward-journald -tag     dock...
           ├─12951 /usr/bin/docker daemon --selinux-enabled -H tcp://0.0.0.0:2376     -H unix:///var/run/docker.sock --tlscacert=/etc/docker/ca.pem     --tlscert=/etc/docker/server-cert.pem --tlskey=/etc/docker/    server-key.pem --tlsverify --storage-driver devicemapper --stora...
           └─12952 /usr/bin/forward-journald -tag docker
```

You are getting among other things the main pid. For the restart you can confirm that the pid changes.

brgnepal: @hferentschik : Updated tests to verify operations being performed.

brgnepal:
@bexelbie @navidshaikh @hferentschik Updated the tests.

locales/en.yml
navidshaikh: > This is imo not consistent with for example vagrant service-manager status

@hferentschik : I am not able to recollect why supporting systemd services are not consistent with `vagrant service-manager status`

README.md
```diff
@@ -104,9 +104,15 @@ The following section lists the available commands for the plugin and their expl
...
 
-   Restarts the given service in the box.
+   where `operation` could be:
+
+   * `restart` : Restart the given service in the box.
+   * `start`   : Start the given service in the box.
+   * `stop`    : Stop the given service in the box.
+
+   and `service` could be `docker`, `openshift`, `kubernetes` or any systemd service.
```
hferentschik: So 'any systems service' is still there?

features/service-operation.feature
hferentschik: Any reason why you are indenting here?

features/service-operation.feature
```diff
@@ -0,0 +1,35 @@
...
+      Then service "docker" should be stopped
+
+      And I successfully run `bundle exec vagrant service-manager start docker`
```
hferentschik, line 27: This is most likely a new 'When'

hferentschik, line 27:
```
When the "docker" service is not running
And I successfully run `bundle exec vagrant service-manager start docker`
Then service "docker" should be running
```

I think this is how I would formulate it

features/service-operation.feature
hferentschik:
```
 And the "docker" service is running
 And I successfully run `bundle exec vagrant service-manager stop docker`
 Then service "docker" should be stopped
```

features/service-operation.feature
```diff
@@ -0,0 +1,35 @@
...
+      Then service "docker" should be started
+
+      Then command "bundle exec vagrant service-manager restart docker" should restart "docker"
```
hferentschik:
```
When the "docker" service is running
And I successfully run `bundle exec vagrant service-manager restart docker`
Then the service "docker" should be running
And have a new pid
```

hferentschik: Most likely this is also a valid case:

```
When the "docker" service is *not* running
And I successfully run `bundle exec vagrant service-manager restart docker`
Then the service "docker" should have restarted
```

features/support/env.rb
```diff
@@ -81,3 +81,27 @@ def output_is_script_readable(raw_stdout)
...
+    regexp = /Active: inactive\(dead\)/
+  end
+
+  expect(last_command_started).to have_exit_status(exit_code)
+  aruba.command_monitor.find(Aruba.platform.detect_ruby(last_command_started)).send(:stdout) =~ regexp
+end
+
+Then(/^command "([^"]*)" should restart "([^"]*)"$/) do |cmd, service|
+  run("vagrant ssh -c \"sudo systemctl status docker\"")
+  pid = last_command_started.pid
```
hferentschik: I don't think this is right at all. The pids you are comparing are the pids of the commands you are executing. These are not the pids of the docker daemon. You need to get the docker daemon pid from the output of systemctl status.

You most likely need to store the pid in a instance variable:

```
When the "docker" service is running
```

The step impl for this would verify that the service is indeed running and store its pid in `@service_pid`.

```
Then the service "docker" should have restarted
```

In the step impl here you would again verify that the servive is running and extract its pid (current_pid). Then you assert @service_pid != current_pid.

lib/vagrant-service-manager/command.rb
```diff
@@ -111,13 +111,13 @@ def execute
...
           end
-        when 'restart'
+        when 'restart', 'start', 'stop'
           exit_if_machine_not_running
           case subcommand
           when '--help', '-h'
             print_help(type: command)
           else
-            restart_service(subcommand)
+            perform_service(command, subcommand)
```
hferentschik, line 119: _perform_service_action_!?

hferentschik, line 119: _subcommand_!? In _perform_service_ the paramter is then called _service_!

lib/vagrant-service-manager/command.rb
hferentschik: -1 for this part. The service-manager is not a proxy to systemctl. How do you explain cohesively the term "service" in the context of vagrant-service-manager when we are including systemd services in this command. How come I can start/stop/restart systemd commands via the plugin, but not get the status of any of these services?

What is wrong with:

```
systemctl start cockpit
```

I am still missing the use case here.

hferentschik:
I am also seeing two test failures, both with the ADB box:

First  in the env_command feature:

```
expected that command "bundle exec vagrant service-manager env openshift" has exit status of "126", but has "0".
```

And second in the status_command fearture:

```
-openshift - stopped
 ...
+Configured services:
+docker - running
+openshift - running
+kubernetes - running
```

The OpenShift service is expected not to be running, but it seems to be nevertheless. 

I am not sure whether I might have an old version of the ADB. I get the latest and re-run the tests.

hferentschik:
Test failures are still there.

README.md
brgnepal: Yes at the moment.

brgnepal:
@hferentschik Probably you have ran the test with ADB 2.0 where there is old `sccli`. ADB latest version 2.1 has new `sccli` - https://github.com/projectatomic/adb-utils/blob/master/utils/sccli.py

hferentschik:
>  Probably you have ran the test with ADB 2.0 where there is old sccli. ADB latest version 2.1 has new sccli - https://github.com/projectatomic/adb-utils/blob/master/utils/sccli.py

In which case the download URL for the get_adb task should probably change.

README.md
hferentschik: Let's wait then to see what the outcome is of the mailing list discussion.

brgnepal: @hferentschik I hope you mean to say. Let's go ahead with this PR merge and later on implement based on ML discussion on systemd service as I think that discussion is out of scope of PR implementation since we already had `restart` implementation for `systemd` services.

hferentschik: > I hope you mean to say. Let's go ahead with this PR merge and later on 
implement based on ML discussion on systemd service.

-1 It is always harder to remove something than adding it. I see me reason 
to add it, if there is a chance we remove it later. In order to merge I'd 
remove any mention of systemd services. Then if we want to go ahead and use 
the plugin also to control systemd services we can add the appropriate 
docs.

hferentschik: > I hope you mean to say. Let's go ahead with this PR merge and later on implement based on ML discussion on systemd service.

-1 It is always harder to remove something than adding it. I see me reason
to add it, if there is a chance we remove it later. In order to merge I'd
remove any mention of systemd services. Then if we want to go ahead and use
the plugin also to control systemd services we can add the appropriate
docs.

hferentschik:
What's about we get the tests implemented as discussed, make sure the full test harness passes and remove the mention of sytemctl for now? Then I think we can merge and move on.

hferentschik:
Using _AtomicDeveloperBundle-2.1.0-CentOS7_ the test pass. So we should update the referenced box in the _Rakefile_.

brgnepal:
Updated with all the changes suggested by Hardy.

@hferentschik , @navidshaikh @LalatenduMohanty @bexelbie @praveenkumar :

Removed support of any kind of operations for systemd services.
We can add it's support based on our discussion through mailing list.

features/support/env.rb
```diff
@@ -81,3 +86,53 @@ def output_is_script_readable(raw_stdout)
...
+
+# Note: Only for services supported through systemctl. Not for 'kubernetes'.
+Then(/^the service "([^"]*)" should be ([^"]*)$/) do |service, operation|
+  run("vagrant ssh -c \"sudo systemctl status #{service}\"")
+
+  if ['running', 'restarted'].include? operation
+    exit_code = 0
+    regexp = /Active: active \(running\)/
+  elsif operation == 'stopped'
+    exit_code = 3
```
navidshaikh: exit codes needs to be referred <https://github.com/projectatomic/vagrant-service-manager#exit-codes> or add if required.

budhrg: These exit codes are result of `systemctl` command and it is different than our VSM exit codes.

navidshaikh: Yup, thanks!

Rakefile
```diff
@@ -10,7 +10,7 @@ CDK_DOWNLOAD_URL='https://access.redhat.com/downloads/content/293/ver=2/rhel---7
 CDK_BOX_BASE_NAME='rhel-cdk-kubernetes-7.2-23.x86_64.vagrant'
 
 ADB_DOWNLOAD_URL='http://cloud.centos.org/centos/7/atomic/images'
-ADB_BOX_BASE_NAME='AtomicDeveloperBundle-2.0.0-CentOS7'
+ADB_BOX_BASE_NAME='AtomicDeveloperBundle-2.1.0-CentOS7'
```
hferentschik, line 13: +1

features/support/env.rb
```diff
@@ -81,3 +86,53 @@ def output_is_script_readable(raw_stdout)
...
+  step "the \"docker\" service is not running"
+end
+
+# Note: Only for services supported through systemctl. Not for 'kubernetes'.
+Then(/^have a new pid for "([^"]*)" service$/) do |service|
+  run("vagrant ssh -c \"sudo systemctl status #{service}\"")
+
+  expect(last_command_started).to have_exit_status(0)
+  stdout = aruba.command_monitor.find(Aruba.platform.detect_ruby(last_command_started)).send(:stdout)
+  expect(@service_current_process_id).not_to eq(extract_process_id(stdout))
```
hferentschik, line 137: :-)

hferentschik:
Merged, thanks.

PR #241, opened 2016-05-31
[WIP] Fix #90: Add install-cli command

### LINKED ISSUE #90: Add install-cli command
In order to get the latest docker|openshift|kubernetes CLI tools to the host, service-manager could offer a `install-cli <service>` command, which depending on the host OS downloads the appropriate CLI tool. 

Since bundling the CLI tools might be a problem in the ADB/CDK VM as well as in the service-manager, the implementation could be based on a download matching CLI version from the respective project download sites.

brgnepal:
Fix for issue #90 - install service cli tools
- Download docker binary for linux ATM

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -1,4 +1,3 @@
-require_relative 'os'
 require 'digest'
 require_relative 'plugin_util'
 require_relative 'plugin_logger'
@@ -17,6 +16,8 @@ module VagrantPlugins
       'docker' => Docker, 'openshift' => OpenShift,
       'kubernetes' => Kubernetes
     }
+    # Download binary at current folder
+    BIN_FOLDER = './'
 
     class Command < Vagrant.plugin(2, :command)
       OS_RELEASE_FILE = '/etc/os-release'
@@ -118,6 +119,16 @@ module VagrantPlugins
           else
             perform_service(command, subcommand)
           end
+        when 'install-cli'
+          exit_if_machine_not_running
+          # Transform hyphen into underscore which is valid char in method
+          command = command.tr('-', '_')
+          case subcommand
+          when '--help', '-h'
+            print_help(type: command)
+          else
+            execute_install_cli(subcommand)
+          end
         when '--help', '-h', 'help'
           print_help
         else
@@ -205,6 +216,19 @@ module VagrantPlugins
       def display_box_ip
         with_target_vms(nil, single_target: true) do |machine|
           @env.ui.info machine.guest.capability(:machine_ip)
+        end
+      end
+
+      def execute_install_cli(service)
+        if service.nil?
+          help_msg = I18n.t('servicemanager.commands.help.install_cli')
+          service_missing_msg = I18n.t('servicemanager.commands.operation.service_missing')
+          @env.ui.error help_msg.gsub(/Install the client side tool for the service/, service_missing_msg)
+          exit 126
+        end
+
+        with_target_vms(nil, single_target: true) do |machine|
+          PluginUtil.service_class(service).install_cli(machine, @env.ui)
         end
       end
     end
--- a/lib/vagrant-service-manager/plugin_util.rb
+++ b/lib/vagrant-service-manager/plugin_util.rb
@@ -94,6 +94,15 @@ module VagrantPlugins
         ui.error e.message.squeeze
       end
 
+      def self.execute_once(machine, ui, command)
+        machine.communicate.sudo(command) do |type, data|
+          PluginLogger.debug
+          return data.chomp
+        end
+      rescue StandardError => e
+        ui.error e.message.squeeze
+      end
+
       def self.print_shell_configure_info(ui, command = '')
         label = if OS.unix?
                   'unix_configure_info'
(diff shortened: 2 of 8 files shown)


brgnepal:
Sending this first cut of implementation.

Expecting feedbacks :smile:  @hferentschik @bexelbie @navidshaikh (in your own time).

lib/vagrant-service-manager/plugin_util.rb
```diff
@@ -117,6 +126,32 @@ def self.env_label(script_readable)
...
+
+      def self.binary_downloaded?(version)
+        File.file?("#{BIN_FOLDER}docker-#{version}")
+      end
+
+      def self.download_binary(ui, url)
+        path = BIN_FOLDER + File.basename(url)
+
+        ui.info 'Downloading binary from url ' + url
+        Vagrant::Util::Downloader.new(url, path).download!
```
hferentschik, line 151: Will this work on Windows, given that Downloader delegates afaik to curl? Why not download it with Ruby native libraries?

hferentschik, line 151: where does the binary go?

hferentschik:
Some thoughts:
- This is just for the docker cli, right? The plan is to support oc, oadm, kubectl as well right
- Where do the binaries go? And how are they then executed by the user
- Should we not update the PATH accordingly, so that the user can use the binary from the current shell? I guess this would require something like the _eval_ approach we have for the _env_ command
- Tests, tests, tests - I'd like to see acceptance tests, but I think it is time to write unit tests as well!

lib/vagrant-service-manager/command.rb
```diff
@@ -17,6 +16,8 @@ module ServiceManager
       'docker' => Docker, 'openshift' => OpenShift,
       'kubernetes' => Kubernetes
     }
+    # Download binary at current folder
+    BIN_FOLDER = './'
```
hferentschik, line 20: not sure about that. I think the default should be something like _bin_ relative to the Vagrantfile, but maybe we want to have an additional option for the install-cli which allows to specify the target directory.

plugins/hosts/linux/cap/os_arch.rb
```diff
@@ -0,0 +1,20 @@
+include Vagrant::Util
+
+module VagrantPlugins
```
hferentschik, line 3: why do we need that?

plugins/hosts/linux/plugin.rb
```diff
@@ -0,0 +1,13 @@
+require 'vagrant'
+
+module VagrantPlugins
```
hferentschik, line 3: And why do we need that?

plugins/hosts/linux/cap/os_arch.rb
brgnepal, line 3: Are you talking about module name `VagrantPlugins`? If yes, for namespace purpose.

lib/vagrant-service-manager/plugin_util.rb
brgnepal, line 151: > Will this work on Windows

No. Will work on it once it got the Windows Hyper-V. Well I doubt Vagrant should have generalized this downloader piece to work in windows too. Would like to see how it perform there and will change my code accordingly.

Right now we discuss to place it to current folder. so that is is easy to call like

```
$ ./docker-1.9.1 images
```

I like suggestion on adding extra option for path where user might want to place binary.

brgnepal:
@hferentschik 

> This is just for the docker cli, right? The plan is to support oc, oadm, kubectl as well right

Yes

> Where do the binaries go? And how are they then executed by the user

Binary goes in current folder relative to Vagrantfile. `--path` would enable to place it in specified path.
User will use as: 

```
$ vagrant service-manager install-cli docker --path /usr/bin
   Downloading binary from url DOCKER_URL
   Binary downloaded at /usr/bin/docker-1.9.1

   Use binary as:
        $ /usr/bin/docker-1.9.1 --help

$  /usr/bin/docker-1.9.1  images
```

**NOTE**: Usage info is printed on output of `install-cli` command.

> Should we not update the PATH accordingly, so that the user can use the binary from the current shell? I guess this would require something like the eval approach we have for the env command

Since, if we update the path `PATH`, we would enable binary to be used from current shell but if user want to use system docker? One of the user reported similar use case in `container-tools` mailing list recently.
I think putting the binaries to current folder along with `--path` is better approach to go at first go IMO.

> Tests, tests, tests - I'd like to see acceptance tests, but I think it is time to write unit tests as well!

On its way :smile:

plugins/hosts/linux/cap/os_arch.rb
hferentschik, line 3: I am talking about the whole class. Trying to understand what's going on here.

brgnepal, line 3: I couldn't find the built-in method to get the architecture info. hence implementing my own here which will be used in multiple places.

hferentschik:
> Binary goes in current folder relative to Vagrantfile

As I said, I'd prefer a sub-directory _bin_, especially once more than one binary gets downloaded. 

Adding some more comments on the issue itself.

plugins/hosts/linux/cap/os_arch.rb
hferentschik, line 3: Ok, I see, you need to further qualify in order to download the right binary. I would think that is provided somewhere, but maybe not.

brgnepal, line 3: > qualify in order to download the right binary 

Didn't get by right binary?

Code to download binary is present in `lib/vagrant-service-manager/plugin_util.rb` line 147.

hferentschik:
Can we close this? AFAIU it was just created for discussion purposes.

brgnepal:
@hferentschik Ok

Issue #230, opened 2016-05-24
Improve acceptance test run time


hferentschik:
This is a follow up on issue #168. The commits added for this issue add a new test file _debug_flag.feature_. The test is fine, but also fits nicely into_env-command.feature_ where already several combinations of the _env_ command are exercised. Adding the debug falg test there will reduce the run time of the tests considerably (4 less startups of the VM).

hferentschik:
Merged.

PR #245, opened 2016-06-02
Issue #230 Moving test from debug-flag.feature into env-command.featu…

hferentschik:
…re to reduce amount of times the VM needs to be started


hferentschik:
Merged

PR #260, opened 2016-06-09
[WIP] Fix #90: Add install-cli command

### LINKED ISSUE #90: Add install-cli command

brgnepal:
Fix #90 
_Only linux supported at present._

Example execution:

```
$ bundle exec vagrant service-manager install-cli docker
# Binary now available at /home/budhram/.vagrant.d/data/service-manager/bin/docker/1.9.1/docker

# export PATH=/home/budhram/.vagrant.d/data/service-manager/bin/docker/1.9.1:$PATH

# run binary as:
# docker <command>

$ export PATH=/home/budhram/.vagrant.d/data/service-manager/bin/docker/1.9.1:$PATH

$ docker version
Client:
 Version:      1.9.1
 API version:  1.21
 Go version:   go1.4.3
 Git commit:   a34a1d5
 Built:        Fri Nov 20 17:56:04 UTC 2015
 OS/Arch:      linux/amd64
Cannot connect to the Docker daemon. Is the docker daemon running on this host?

$ bundle exec vagrant service-manager install-cli docker                          
# Binary already available at /home/budhram/.vagrant.d/data/service-manager/bin/docker/1.9.1/docker

# export PATH=/home/budhram/.vagrant.d/data/service-manager/bin/docker/1.9.1:$PATH

# run binary as:
# docker <command>
```

Few points to note:
- One can use system binary by **opening a new terminal tab/window**.
- In last command, 'already' word is written for same version of binary
- Added `--path` to specify destination binary path
- Added `--cli-version` to specify binary version. `--version` conflicting with vagrant `--version` flag.

Not adding `--url` as it is very tricky and have no control over user's input.

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -1,7 +1,9 @@
-require_relative 'os'
 require 'digest'
 require_relative 'plugin_util'
 require_relative 'plugin_logger'
+require_relative 'installer'
+require_relative 'adb_installer'
+require_relative 'cdk_installer'
 
 module VagrantPlugins
   module ServiceManager
@@ -17,6 +19,14 @@ module VagrantPlugins
       'docker' => Docker, 'openshift' => OpenShift,
       'kubernetes' => Kubernetes
     }
+
+    def self.bin_dir
+      @bin_dir
+    end
+
+    def self.set_bin_dir(env)
+      @bin_dir = "#{env.data_dir}/service-manager/bin"
+    end
 
     class Command < Vagrant.plugin(2, :command)
       OS_RELEASE_FILE = '/etc/os-release'
@@ -37,6 +47,8 @@ module VagrantPlugins
       end
 
       def execute
+        ServiceManager.set_bin_dir(@env)
+
         argv = ARGV.dup
         # Don't propagate --debug argument to case operation
         if ARGV.include? '--debug'
@@ -120,6 +132,19 @@ module VagrantPlugins
           else
             perform_service(command, subcommand)
           end
+        when 'install-cli'
+          exit_if_machine_not_running
+          # Transform hyphen into underscore which is valid char in method
+          command = command.tr('-', '_')
+          # Get options as Hash which are preceeded with -- like --version
+          options = Hash[ARGV.drop(3).each_slice(2).to_a]
+
+          case subcommand
+          when '--help', '-h'
+            print_help(type: command)
+          else
+            execute_install_cli(subcommand, options)
+          end
         when '--help', '-h', 'help'
           print_help
         else
@@ -129,7 +154,7 @@ module VagrantPlugins
 
       def execute_service(name, options = {})
         with_target_vms(nil, single_target: true) do |machine|
-          PluginUtil.service_class(name).info(machine, @env.ui, options)
+          PluginUtil.service_class(name).new(machine, @env).info(options)
         end
       end
 
@@ -144,11 +169,11 @@ module VagrantPlugins
       def execute_status_display(service = nil)
         with_target_vms(nil, single_target: true) do |machine|
           if service && SUPPORTED_SERVICES.include?(service)
-            PluginUtil.service_class(service).status(machine, @env.ui, service)
+            PluginUtil.service_class(service).new(machine, @env).status
           elsif service.nil?
             @env.ui.info I18n.t('servicemanager.commands.status.nil')
             SUPPORTED_SERVICES.each do |s|
-              PluginUtil.service_class(s).status(machine, @env.ui, s)
+              PluginUtil.service_class(s).new(machine, @env).status
             end
           else
             @env.ui.error I18n.t('servicemanager.commands.status.unsupported_service',
@@ -171,7 +196,7 @@ module VagrantPlugins
             end
             # since we do not have feature to show the Kube connection information
             unless service == 'kubernetes'
-              service_class.info(machine, @env.ui, options)
+              service_class.new(machine, @env).info(options)
             end
           end
 
@@ -214,6 +239,34 @@ module VagrantPlugins
 
         with_target_vms(nil, single_target: true) do |machine|
           @env.ui.info machine.guest.capability(:machine_ip, options)
+        end
+      end
+
+      def execute_install_cli(service, options = {})
+        if service.nil?
+          help_msg = I18n.t('servicemanager.commands.help.install_cli')
+          service_missing_msg = I18n.t('servicemanager.commands.operation.service_missing')
+          @env.ui.error help_msg.gsub(/Install the client side tool for the service/, service_missing_msg)
+          exit 126
+        end
+
+        with_target_vms(nil, single_target: true) do |machine|
+          installer = ADBInstaller
+          args = [service.to_sym, machine, @env, options]
+
+          begin
+            box_version = machine.guest.capability(:os_variant)
+
+            if box_version == 'cdk'
+              ui.info I18n.t('servicemanager.commands.install_cli.unsupported_box')
+              exit 126
+              # TODO: installer = CDKInstaller.new
+            end
+            installer.new(*args).install
+          rescue Vagrant::Errors::GuestCapabilityNotFound
+            ui.info 'Not a supported box.'
+            exit 126
+          end
         end
       end
     end
--- a/lib/vagrant-service-manager/installer.rb
+++ b/lib/vagrant-service-manager/installer.rb
@@ -1,0 +1,55 @@
+module VagrantPlugins
+  module ServiceManager
+    class Installer
+      LABEL = 'servicemanager.commands.install_cli.message'
+      TEMP_DIR = '/tmp'
+      BINARY_NAME = {
+        docker: 'docker', openshift: 'oc'
+      }
+
+      def initialize(type, machine, env, options)
+        @type = type
+        @machine = machine
+        @ui = env.ui
+        @options = options
+
+        @binary_exists = true
+        @version = options['--cli-version'] || PluginUtil.execute_once(@machine, @ui, VERSION_CMD[@type])
+        @path = options['--path']
+        @binary_name = BINARY_NAME[@type]
+        @temp_bin_dir = "#{TEMP_DIR}/#{@type.to_s}"
+        @url = ''
+      end
+
+      def install
+        puts "I m install now...."
+        build_binary_path
+
+        unless PluginUtil.binary_downloaded?(@path)
+          build_download_url
+          puts "url: #{@url}"
+          ensure_binary_and_temp_directories
+          download_and_prepare_binary
+        end
+      end
+
+      def print_message
+        @ui.info I18n.t(LABEL, path: @path, dir: File.dirname(@path),
+                        binary: @binary_name, when: (@binary_exists ? 'already' : 'now'))
+      end
+
+      def build_binary_path
+        @path = "#{ServiceManager.bin_dir}/#{@type.to_s}/#{@version}/#{@binary_name}" if @path.nil?
+      end
+
+      def build_download_url
+      end
+
+      def ensure_binary_and_temp_directories
+      end
+
+      def download_and_prepare_binary
+      end
+    end
+  end
+end
--- a/lib/vagrant-service-manager/plugin_util.rb
+++ b/lib/vagrant-service-manager/plugin_util.rb
@@ -94,6 +94,15 @@ module VagrantPlugins
         ui.error e.message.squeeze
       end
 
+      def self.execute_once(machine, ui, command)
+        machine.communicate.sudo(command) do |type, data|
+          PluginLogger.debug
+          return data.chomp
+        end
+      rescue StandardError => e
+        ui.error e.message.squeeze
+      end
+
       def self.print_shell_configure_info(ui, command = '')
         label = if OS.unix?
                   'unix_configure_info'
@@ -116,6 +125,10 @@ module VagrantPlugins
         else
           'windows'
         end
+      end
+
+      def self.binary_downloaded?(path)
+        File.file?(path)
       end
     end
   end
--- a/lib/vagrant-service-manager/services/docker.rb
+++ b/lib/vagrant-service-manager/services/docker.rb
@@ -1,13 +1,13 @@
 module VagrantPlugins
   module ServiceManager
-    class Docker
+    class Docker < ServiceBase
       # Hard Code the Docker port because it is fixed on the VM
       # This also makes it easier for the plugin to be cross-provider
       PORT = 2376
 
-      def initialize(machine, ui)
-        @machine = machine
-        @ui = ui
+      def initialize(machine, env)
+        super(machine, env)
+        @service_name = 'docker'
       end
 
       def execute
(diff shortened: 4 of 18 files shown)


lib/vagrant-service-manager/command.rb
```diff
@@ -17,6 +17,8 @@ module ServiceManager
       'docker' => Docker, 'openshift' => OpenShift,
       'kubernetes' => Kubernetes
     }
+    # Download binary at current folder
+    BIN_FOLDER = "#{ENV['HOME']}/.vagrant.d/data/service-manager/bin"
```
hferentschik: ENV['HOME'] is not going to work on Windows in all cases. The Vagrant environment contains the data dir. You should extract it from there

hferentschik:
> Using format binary name + version in order to avoid conflict with system client binaries.

-1 I don't want to have to think about which docker version I need to call. The whole point of having this command is that the requested version is the one being called in the configured shell.

>  Another important reason is one couldn't able to export same binary name which is already available in system. Example: I was trying to export for docker binary but getting system binary even export command is successful.

Hmm, looking at your export command this is not surprising 

```
export PATH=$PATH:/home/budhram/.vagrant.d/data/service-manager/bin/docker/1.9.1
```

You are adding the binary at the end of the path. Of course a potential system binary gets precedence in this case. It should be more like

   export PATH=/home/budhram/.vagrant.d/data/service-manager/bin/docker/1.9.1:$PATH

You need to make sure that your binary is first on the path.

lib/vagrant-service-manager/command.rb
brgnepal: Noted. Thanks.

brgnepal:
@hferentschik I am having less knowledge of shell scripting here. 
Thanks to your suggestion of setting path before `$PATH`. 

I will update the description now.

lib/vagrant-service-manager/command.rb
hferentschik: So are you going to  fix it? It is either part of the @env instance you are getting passed - https://github.com/vagrant-landrush/landrush/blob/master/lib/landrush/command.rb#L11 or you need to use the home_path - https://github.com/vagrant-landrush/landrush/blob/master/lib/landrush/action/setup.rb#L9. You need to test.

lib/vagrant-service-manager/command.rb
```diff
@@ -120,6 +122,24 @@ def execute
...
             perform_service(command, subcommand)
           end
+        when 'install-cli'
+          exit_if_machine_not_running
+          # Transform hyphen into underscore which is valid char in method
+          command = command.tr('-', '_')
+          # Get options as Hash which are preceeded with -- like --version
+          options = ARGV.drop(3).each_slice(2).to_a.to_h
+
+          if options['--url'] && options['--cli-version'].nil?
```
hferentschik: I thought there is no --url option for now?

lib/vagrant-service-manager/installer.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'adb_executor'
+require_relative 'cdk_executor'
+
+module VagrantPlugins
+  module ServiceManager
+    module Executor
+      def self.build(type, machine, ui, options)
```
hferentschik: All this logic is testable without going all the way with the starting a new VM. I think we need to introduce unit tests now. Here is a commit - https://github.com/hferentschik/adb-vagrant-registration/commit/18679a122c11c03072ee63617c48e67f2e97bb39 - which introduces minitest for vagrant-registration. It is a very simple setup and builds on mocking out some of the core classes of Vagrant. This way you can execute commands, capabilities, etc without going end to end. For example a test could make sure that the right executor is returned. In test you basically record which response the os_variant is supposed to return and then you can assert what Executor is returned. You can also test what happens when the variant is not at all adb or cdk. Just as an example.

lib/vagrant-service-manager/plugin_util.rb
```diff
@@ -94,6 +94,15 @@ def self.execute_and_exit_on_fail(machine, ui, command)
         ui.error e.message.squeeze
       end
```
hferentschik, line 96: I find PluginUtil a bit odd. It starts getting very long and seems to be a class for everything. I think there are more than one class which are needed there.

lib/vagrant-service-manager/adb_executor.rb
```diff
@@ -0,0 +1,40 @@
...
+        version = PluginUtil.execute_once(@machine, @ui, VERSION_CMD[@type.to_sym]) if version.nil?
+        path = PluginUtil.service_class(@type).binary_path(version) if path.nil?
+
+        unless PluginUtil.binary_downloaded?(path)
+          binary_exists = false
+          url = PluginUtil.service_class(@type).binary_download_url(@machine, version)
+          PluginUtil.download_binary(@type, url, path)
+        end
+
+        binary_name = @type
```
hferentschik: These two lines are just confusing

lib/vagrant-service-manager/adb_executor.rb
```diff
@@ -0,0 +1,40 @@
...
+        @ui = ui
+        @options = options
+      end
+
+      def execute
+        binary_exists = true
+        version = @options['--cli-version']
+        path = @options['--path']
+
+        version = PluginUtil.execute_once(@machine, @ui, VERSION_CMD[@type.to_sym]) if version.nil?
```
hferentschik: Why is PluginUtil not a mixin of ADBExecutor?

lib/vagrant-service-manager/services/open_shift.rb
```diff
@@ -60,6 +63,30 @@ def self.print_info(ui, options)
...
+        download_base_path = "#{BINARY_BASE_URL}/download/v#{version}/"
+        file = "#{FILE_PREFIX}-v#{version}-#{sha}-#{os_type}-#{ext(machine, os_type)}"
+        download_base_path + file
+      end
+
+      def self.download_binary(url, path)
+        Dir.mkdir(TEMP_DIR) unless File.directory? TEMP_DIR
+        Vagrant::Util::Downloader.new(url, "#{TEMP_DIR}/#{File.basename(url)}").download!
+        dir = File.basename(url).sub('.tar.gz', '')
+        system("tar -xf #{TEMP_DIR}/#{File.basename(url)} -C #{TEMP_DIR}")
```
hferentschik: this might not work on Windows. It is not a good idea to delegate to system command. Ruby has built-in classes for this - http://stackoverflow.com/questions/856891/unzip-zip-tar-tag-gz-files-with-ruby

lib/vagrant-service-manager/services/open_shift.rb
```diff
@@ -60,6 +63,30 @@ def self.print_info(ui, options)
         end
       end
 
+      def self.install_cli(machine, ui, options)
+        Installer.new('openshift', machine, ui, options).run
```
hferentschik: instead of string for the services, should you not use symbols for our known services? :docker, :openshift, :kubernetes

lib/vagrant-service-manager/services/open_shift.rb
```diff
@@ -60,6 +63,30 @@ def self.print_info(ui, options)
...
+        arch = machine.env.host.capability(:os_arch)
+        sha = version_sha(version)
+        download_base_path = "#{BINARY_BASE_URL}/download/v#{version}/"
+        file = "#{FILE_PREFIX}-v#{version}-#{sha}-#{os_type}-#{ext(machine, os_type)}"
+        download_base_path + file
+      end
+
+      def self.download_binary(url, path)
+        Dir.mkdir(TEMP_DIR) unless File.directory? TEMP_DIR
+        Vagrant::Util::Downloader.new(url, "#{TEMP_DIR}/#{File.basename(url)}").download!
```
hferentschik: Have you checked whether the Downloader works on Windows?

hferentschik:
I think we are on the right track. Getting there. Right now I would like to see a unit test framework though and some simple unit tests. We have been putting this off for too long. I think thinking about how to test certain things will also help making some different choices on how to group functions.

There are also a couple of things I am still unsure about when it comes to Windows. I'd like to see this tested on Windows as well.

lib/vagrant-service-manager/services/open_shift.rb
```diff
@@ -60,6 +63,30 @@ def self.print_info(ui, options)
...
+      def self.install_cli(machine, ui, options)
+        Installer.new('openshift', machine, ui, options).run
+      end
+
+      def self.binary_path(version)
+        "#{BIN_FOLDER}/openshift/#{version}/oc"
+      end
+
+      def self.binary_download_url(machine, version)
+        arch = machine.env.host.capability(:os_arch)
```
hferentschik: Is this not exactly what we wanted to avoid? Now this service class knows how to download for ADB. What's about the CDK? Was the idea not to encapsulate this?

lib/vagrant-service-manager/services/docker.rb
```diff
@@ -71,6 +69,35 @@ def self.print_env_info(ui, options)
...
+        Installer.new('docker', machine, ui, options).run
+      end
+
+      def self.binary_path(version)
+        "#{BIN_FOLDER}/docker/#{version}/docker"
+      end
+
+      def self.binary_download_url(machine, version)
+        arch = machine.env.host.capability(:os_arch)
+        "#{BINARY_BASE_URL}/#{os_type}/#{arch}/docker-#{version}"
```
hferentschik: Is this not exactly what we wanted to avoid? Now this service class knows how to download for ADB. What's about the CDK? Was the idea not to encapsulate this?

lib/vagrant-service-manager/installer.rb
brgnepal: @hferentschik Ya we need to add unit tests now.

lib/vagrant-service-manager/adb_executor.rb
brgnepal: @hferentschik Actually, binary name is different then the type like in case of openshift it is oc and kubectl in case of kubernetes but for docker it is docker. 

May be we can pull this code into `PluginUtil`.

lib/vagrant-service-manager/command.rb
brgnepal: Yes next iteration :)

lib/vagrant-service-manager/command.rb
brgnepal: Ya forgot to remove it :)

lib/vagrant-service-manager/services/docker.rb
brgnepal: I don't think soo. This is just a download url method and I believe each service class should know its download url.
For CDK, this logic will be updated as conditional based generation of url.

lib/vagrant-service-manager/services/open_shift.rb
brgnepal: Agreed +1

lib/vagrant-service-manager/services/open_shift.rb
brgnepal: Not yet. Will be in next iteration.

lib/vagrant-service-manager/services/open_shift.rb
brgnepal: Ya sure

lib/vagrant-service-manager/adb_executor.rb
brgnepal: You mean to say include `PluginUtil` inside `ADBExecutor` class?

I don't feel good including whole module here as I only need two or three methods

lib/vagrant-service-manager/command.rb
hferentschik: > next iteration

what does that mean? Which iteration are you talking about?

lib/vagrant-service-manager/command.rb
hferentschik: ok

lib/vagrant-service-manager/installer.rb
hferentschik: +1, I suggest you align with what I am doing in vagrant-registration. This was we are consistent between plugins and it is easier to understand/extend.

lib/vagrant-service-manager/adb_executor.rb
hferentschik: > Actually, binary name is different then the type like in case of openshift it is oc and kubectl in case of kubernetes but for docker it is docker.

I understand what you are doing, but it is not really easy to understand. I think there are several things which could improve that. The easiest being a dedicated method with a descriptive name/. Or maybe one of the helper methods above can return the name?

> May be we can pull this code into PluginUtil.

I am not a big fan of this idea. As said, `PluginUtil` is imo already too big and dealing with too many different aspects.

lib/vagrant-service-manager/adb_executor.rb
hferentschik: `PluginUtil` is a helper class right? By including it you can call these helper methods directly without always having to qualify `PluginUtil`. This is just a readability thing. IMO the code becomes more concise. Is that not the Ruby way?

lib/vagrant-service-manager/services/docker.rb
hferentschik: But you will end up with switch statement in the service. Exactly what we tried to avoid. In this case  why did you create these executors in the first place. Also a `service` imo does not have to have knowledge over where client libs are coming from. And how do you know that the CDK will have a download URL. The mechanism might be completely different in which case things get even more complicated. The how of getting the cli libs should completely be encapsulated in the executors. Your abstractions are leaking.

lib/vagrant-service-manager/adb_executor.rb
brgnepal: Its a module so we need to invoke its method using module name. The way you describe could be achieved by converting into class and making other class subclass of it which I don't think good idea.
For example, think about scenarios where I have to use only one or two method of this plugin util class in `plugins/guests/cap` files. There it is not point to make it super class.

One thing we can do is we can filter few common methods which should be extracted into base class which can be subclass within the `lib/` folder appropriate classes.

lib/vagrant-service-manager/services/docker.rb
brgnepal: > And how do you know that the CDK will have a download URL. The mechanism might be completely different in which case things get even more complicated. 

Probably we need to think it when there is requirement of implementing for CDK. Implementing with wild assumption is what I want to avoid here. At least, the code here assumes that there will be some download URL for CDK and hence I make it separate implementation class for CDK.

hferentschik:
> Its a module so we need to invoke its method using module name. 

No we don't. I am talking about _include_ (mixins), not _extend_. I agree using inheritance for this would be wrong. See also http://www.railstips.org/blog/archives/2009/05/15/include-vs-extend-in-ruby/

hferentschik:
> Probably we need to think it when there is requirement of implementing for CDK. 

And when time comes you have to change code at several places/classes which is a sign of poor design. The whole idea is to just change one class when time comes. 

> Implementing with wild assumption is what I want to avoid here.

I am not making any assumptions at all. It is you who for example assumes that there will be an URL to download from.

hferentschik:
While reviewing the code I found something else which troubles me. Let's take VagrantPlugins::ServiceManager::OpenShift as example. It has an _initialize_ and _execute_ method. So you are creating instances of this service. This happens in VagrantPlugins::ServiceManager::Service. 

All other methods of OpenShift are class methods and are called statically via the VagrantPlugins::ServiceManager::PluginUtil.service_class lookup. The latter looks up a class and then calls class methods. 

I think it really should be one or the other here. Either all is stateless and we make class method class or we instantiate service instances and that's also what we look up via PluginUtil. The benefit of the latter would be that one would not need to pass the machine instance all the time, since it would be a member instance of the service.

hferentschik:
retest this please

brgnepal:
> I am not making any assumptions at all. It is you who for example assumes that there will be an URL to download from.

@hferentschik At least to implement the first cut of it we have to believe that this is our way :smile:

hferentschik:
> At least to implement the first cut of it we have to believe that this is our way

I am not arguing against using a URL to download the CLI binaries for ADB. That fine. I am arguing how the code is structured. I don't believe this is proper OO design. This is also paired with the fact that we mix instance vs class invocations.

brgnepal:
> This is also paired with the fact that we mix instance vs class invocations.

@hferentschik Agreed. Thinking on redesigning now.

LalatenduMohanty:
On OS X I am getting below error

``` r
$ vagrant service-manager install-cli
Service name missing

Usage: vagrant service-manager install-cli [service] [options]

Service:
    A service provided by sccli. For example:
     docker, kubernetes, openshift etc

Options:
      -h, --help         print this help

Example:
      vagrant service-manager install-cli docker
The capability 'os_arch' could not be found. This is an internal error
that users should never see. Please report a bug.
```

LalatenduMohanty:
@budhrg I do not see debug messages in the PR, do you have plan to add some debug messages , so that it will help future debugging?

hferentschik:
I think it would make sense to close this pull request for now and create a new one once all issues we discussed are addressed. @budhrg WDYT?

brgnepal:
@hferentschik Ok sure

PR #265, opened 2016-06-13
No need to make env call evaluable

hferentschik:


hferentschik:
Hmm, not a good idea :-(

PR #269, opened 2016-06-15
Fix #268 Explicitly setting v1.8.4 as dev version for Vagrant. 

### LINKED ISSUE #268: Incompatible version of listen gem
When installing the bundle environment the following error occurs using Ruby 2.x:

```
Installing listen 3.1.5

Gem::InstallError: listen requires Ruby version >= 2.2.3, ~> 2.2.
Installing cucumber 2.4.0
Installing domain_name 0.5.20160310
Installing fog-json 1.0.2
Installing fog-xml 0.1.2
Installing winrm 1.8.1
An error occurred while installing listen (3.1.5), and Bundler cannot continue.
Make sure that `gem install listen -v '3.1.5'` succeeds before bundling.
Connection to n4.pufty.ci.centos.org closed.
Executing: ssh -t -t -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@n4.pufty.ci.centos.org      "yum -y install git epel-release ansible1.9 && yum -y install ansible1.9 && git clone https://github.com/    projectatomic/vagrant-service-manager.git && cd vagrant-service-manager && git checkout master && cd .ci/    ansible && ANSIBLE_NOCOLOR=1 ansible-playbook site.yml"
Executing: ssh -t -t -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@n4.pufty.ci.centos.org      "cd vagrant-service-manager && gem install bundler && bundle install --no-color"
Unable to setup Ruby environment
Build step 'Execute Python script' marked build as failure
Finished: FAILURE
```

The listen 3.1.x versions require Ruby 2.2, but we are using Ruby 2.0 (as does Vagrant afaiu). The listen version needs to be constrained in the _Gemfile_.

hferentschik:
HEAD of Vagrant 1.8.4.dev has switched to use Ruvy 2.2 (2 days ago, see https://github.com/mitchellh/vagrant/blob/master/vagrant.gemspec). I don't think we should follow suite as yet.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,10 +3,12 @@ source 'https://rubygems.org'
 gemspec
 
 group :development do
-  gem 'vagrant',      git: 'https://github.com/mitchellh/vagrant.git'
+  gem 'vagrant',
+      :git => 'git://github.com/mitchellh/vagrant.git',
+      :ref => 'v1.8.4'
   gem 'rake'
-  gem 'vagrant-libvirt'  if RUBY_PLATFORM =~ /linux/i
-  gem 'fog-libvirt', '0.0.3'  if RUBY_PLATFORM =~ /linux/i # https://github.com/pradels/vagrant-libvirt/issues/568
+  gem 'vagrant-libvirt'              if RUBY_PLATFORM =~ /linux/i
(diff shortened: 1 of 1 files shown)


hferentschik:
Looking good. I think this needs to be merged asap.

brgnepal:
LGTM. It seems that now vagrant is using [bundler `1.12.5`](https://github.com/mitchellh/vagrant/blob/master/vagrant.gemspec#L23)

PR #273, opened 2016-06-20
Fix #272: README fix for env default behavior

### LINKED ISSUE #272: service-manager makes docker a default provider?
When I don't provide object name in front of `env` service-manager takes `docker` as default, rather than erroring out or telling me to specify specific provider.

``` bash
$ vagrant service-manager env 

# docker env:
export DOCKER_HOST=tcp://172.28.128.164:2376
[SNIP]
export DOCKER_TLS_VERIFY=1

# eval "$(vagrant service-manager env)"
```

Version:

``` bash
vagrant-service-manager (1.1.0)
```

brgnepal:
Fix #272

Updated `README` to document default behavior of `env` command without service name.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -106,7 +106,9 @@ Displays connection information for all active services in the box in a
 manner that can be evaluated in a shell. If a `service` is specified,
 only the information for that service is displayed. When
 `--script-readable` is specified the output is in `key=value` format.
-The supported services are: Docker; OpenShift.
+The supported services are: Docker, OpenShift.
++
+*Note:* If a `service` is not specified, it displays information of all active services.
 
 1.  `vagrant service-manager box [command]`
 +


README.adoc
```diff
@@ -107,6 +107,8 @@ manner that can be evaluated in a shell. If a `service` is specified,
 only the information for that service is displayed. When
 `--script-readable` is specified the output is in `key=value` format.
 The supported services are: Docker; OpenShift.
```
navidshaikh: @budhrg : although this should not be part of the PR, but while reviewing, I noticed following line can be improved

```
The supported services are: Docker; OpenShift.
```

to 

```
The supported services are: Docker, OpenShift.
```

brgnepal:
@navidshaikh Updated PR

PR #276, opened 2016-06-20
Fix #274: Fixes displaying help irrespective of machine status

### LINKED ISSUE #274: `vagrant service-manager env --help` requires a running machine
$subject, which should not be the case. Help command execution path should be irrespective of the machine running status.

navidshaikh:
Fixes #274

  For all supported commands, -h / --help is now executable irrespective
  of the machine running status.

### DIFF
--- a/lib/vagrant-service-manager/command.rb
+++ b/lib/vagrant-service-manager/command.rb
@@ -49,13 +49,14 @@ module VagrantPlugins
         command, subcommand, option = argv.drop(1)
         case command
         when 'env'
-          exit_if_machine_not_running
           case subcommand
           when 'docker', 'openshift'
             case option
             when nil
+              exit_if_machine_not_running
               execute_service(subcommand)
             when '--script-readable'
+              exit_if_machine_not_running
               execute_service(subcommand, script_readable: true)
             when '--help', '-h'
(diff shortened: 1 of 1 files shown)


brgnepal:
This code is not following Ruby's main principle `DRY`(don't repeat yourself) :smile: .
I definitely feel refactoring here.

cc @hferentschik

hferentschik:
> This code is not following Ruby's main principle

+1 I think there must be a better solution. This is super error prone. Maybe time to break up this massive switch statement? Needs some analysis on what the best approach is

Also this one is perfect for cucumber tests, since we don't need to start the VM in this case. As you guys say, this should work without the VM being up. So we could combine all help tests into the help-command feature and ensure that they print the right content and that they execute without the VM started. With this test in place we won't get a regression on this one.

brgnepal:
@navidshaikh If this is not priority issue then I would like to take this and will come up with better solution for never-ending `execute` method.

But my refactoring may take some time :smile: and may go through iterations based on feedbacks.

navidshaikh:
> @navidshaikh If this is not priority issue then I would like to take this

Sure!

hferentschik:
+1

PR #305, opened 2016-07-08
Fix #281 Making download of box file a true dependency of the feature…

### LINKED ISSUE #281: The ADB/CDK download rake tasks should be made proper Rake tasks
The idea is that the _features_ task can depend on the download tasks and automatically downloads them when the tests are run and the VM's are out of date.

hferentschik:
… task with update checks
- Fixing broken Cucumber test for CDK
- Enabling testing against CDK on CI

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -224,14 +224,8 @@ acceptance tests. They can be run via:
 $ bundle exec rake features
 ---------------------------
 
-*Note:* These tests can be executed in Linux. Not all of the tests run
-appropriately in Windows, pending resolution of
+*Note:* This Cucumber tests do not run on Windows, pending resolution of
 https://github.com/projectatomic/vagrant-service-manager/issues/213[Issue #213].
-
-The tests assume that the ADB and CDK box files are available under
-_build/boxes/adb-<provider>.box_ and
-_build/boxes/cdk-<provider>.box_ respectively. You can either copy the box files
-manually or use the _get_adb_ and _get_cdk_ Rake tasks, respectively.
 
 Per default, only the scenarios for ADB in combination with the
 VirtualBox provider are run. However, you can also run the tests against CDK
@@ -244,7 +238,18 @@ $ bundle exec rake features BOX=cdk PROVIDER=libvirt
 
 # Run against ADB and CDK (boxes are comma separated)
 $ bundle exec rake features BOX=cdk,adb
+
+# Run against ADB and CDK using VirtualBox and Libvirt
+$ bundle exec rake features BOX=cdk,adb PROVIDER=libvirt,virtualbox
 -----------------------------------------------------
+
+The _features_ task will transparently download the required Vagrant
+boxes and cache them in _.boxes_. The cache can be cleared
+via the _clean_boxes_ task.
+
+Using the variable _NIGHTLY=true_ you can make sure that the
+latest nightly build of the CDK is used (VPN access required).
+Per default the latest public release of the CDK is used.
 
 You can also run a single feature specifying the explicit feature file
 to use:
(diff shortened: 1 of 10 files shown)


README.adoc
```diff
@@ -244,8 +238,19 @@ $ bundle exec rake features BOX=cdk PROVIDER=libvirt
 
 # Run against ADB and CDK (boxes are comma separated)
 $ bundle exec rake features BOX=cdk,adb
+
+# Run against ADB and CDK using VirtualBox and Libvirt
+$ bundle exec rake features BOX=cdk,adb PROVIDER=libvirt,virtualbox
 -----------------------------------------------------
 
+The _features_ task will transparently download the required Vagrant
+boxes and cache them in _.boxes_. The cache can be cleared
```
brgnepal, line 247: ".boxes" confused me. could we have path as "./build/boxes" ?

Rakefile
```diff
@@ -5,29 +5,24 @@ require 'mechanize'
...
 require 'launchy'
-
-CDK_DOWNLOAD_URL='https://access.redhat.com/downloads/content/293/ver=2/rhel---7/2.0.0/x86_64/product-software'
-CDK_BOX_BASE_NAME='rhel-cdk-kubernetes-7.2-23.x86_64.vagrant'
-
-CDK_DOWNLOAD_URL_NIGHTLY='http://cdk-builds.usersys.redhat.com/builds/nightly/latest-build'
-
-ADB_DOWNLOAD_URL='http://cloud.centos.org/centos/7/atomic/images'
-ADB_BOX_BASE_NAME='AtomicDeveloperBundle-2.1.0-CentOS7'
+require_relative 'download_task'
```
brgnepal, line 8: :+1:

README.adoc
hferentschik, line 247: no, you want to be outside the _build_ directory, since this gets deleted and then you loose the cached boxes. 'clean' should delete the whole build directory. We did this wrong so far.

download_task.rb
```diff
@@ -0,0 +1,206 @@
...
+require 'date'
+
+class DownloadTask
+  attr_reader :provider
+  attr_reader :box_dir
+  attr_reader :agent
+  attr_reader :meta
+
+  def initialize(box_dir, provider)
+    unless File.directory?(box_dir)
```
brgnepal, line 11: This could have bee in one line. Rubocop would have suggest that :wink:

Rakefile
hferentschik, line 8: I thought you would like that :-)
The Rakefile is then focused on the actual build tasks.

download_task.rb
hferentschik, line 11: :-) Let's cross this bridge once your pull request comes

download_task.rb
```diff
@@ -0,0 +1,206 @@
...
+
+    @meta_file_name = File.join(box_dir, "#{name}-#{provider}.yml")
+    @meta = read_meta
+
+    @agent = Mechanize.new
+    @agent.follow_meta_refresh = true
+    @agent.ignore_bad_chunking = true
+  end
+
+  def needed?
```
brgnepal, line 25: Why it is always true? Does it means, every time I run acceptance test, it will download it?

brgnepal, line 25: Got it.

hferentschik, line 25: This is the base class. There are actually no instances of this. The sub classes override this accordingly. But a safe default is 'true'.

download_task.rb
```diff
@@ -0,0 +1,206 @@
+require 'fileutils'
```
brgnepal, line 1: I am not able to digest this ruby file wandering in root path. Can we move to some folder? 

What about `features/support` folder? Since it being used for that purpose.

hferentschik:
uperseded by pull request #308

PR #315, opened 2016-07-11
Fix #312: message for kubernetes install-cli

### LINKED ISSUE #312: install-cli commands throws an error when trying to install Kubernetes CLI
kubernetes throws error with vsm install-cli 

Steps to reproduce 
- Install latest gem build from master
- run `vagrant.exe service-manager install-cli kubernetes`

Error log  https://gist.github.com/naina-verma/8f82ffa0e0f0ee7261ad62063bb09755
if kubernetes support is not ready then it should not be shown in install-cli help.

brgnepal:
Fix #312 

Since, kubernetes access from host is not supported at present hence we are showing warning for respective install-cli command.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -127,20 +127,17 @@ If existing certificates are found to be invalid, they get regenerated.
 +
 Displays box related information like release version, IP etc.
 +
---
 The possible options for `command` are:
 
 * `version`: Displays the version and release information of the running VM.
 * `ip`: Displays the routable IP address of the running VM.
 +
---
 When `--script-readable` is specified the output is in `key=value` format.
 
 1.  `vagrant service-manager [operation] [service]`
 +
 Manages the life cycle of a service.
 +
---
 The possible options for `operation` are:
 
   * `status`: Lists services and their running state. If a `service` is specified only
@@ -150,10 +147,18 @@ only supported orchestrators are reported.
   * `stop`: Stop the given service in the box.
   * `restart`: Restart the given service in the box.
 +
---
 The supported options for `service` are: Docker, OpenShift.
 
+1. `vagrant service-manager install-cli [service] [--cli-version] [--path]`
++
+Install the client binary for the specified `service`.
++
+The possible options are:
 
+  * `--cli-version`: Download the binary in the specified version, if it exist.
+  * `--path`: Download the binary to the specified path on the host.
++
+The supported options for `service` are: Docker, OpenShift.
 
 [[debug-flag]]
 ==== Debug Flag
--- a/locales/en.yml
+++ b/locales/en.yml
@@ -186,6 +186,8 @@ en:
           Please visit access.redhat.com to download client binaries.
         service_not_enabled: |-
           '%{service}' service is not enabled.
+        kube_not_supported: |-
+          Installation of Kubernetes client library via the install-cli command is not supported yet.
 
       status:
         nil: |-
--- a/test/vagrant-service-manager/installer_test.rb
+++ b/test/vagrant-service-manager/installer_test.rb
@@ -71,6 +71,17 @@ module VagrantPlugins
           @installer.instance_variable_get('@binary_handler').binary_exists.must_equal true
         end
       end
+
+      describe 'Kubernetes' do
+        before do
+          @type = :kubernetes
+          @options = { box_version: 'adb' }
+        end
+
+        it 'should exit' do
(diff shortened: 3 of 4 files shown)


locales/en.yml
```diff
@@ -186,6 +186,10 @@ en:
           Please visit access.redhat.com to download client binaries.
         service_not_enabled: |-
           '%{service}' service is not enabled.
+        kube_not_supportted: |-
+          Client binary installs for kubernetes service is not supported yet.
+          Please refer http://kubernetes.io/docs/getting-started-guides/docker/#download-kubectl and
+          download the appropriate binary.
```
navidshaikh: Even if user downloads the binary, we do not have support for out of the box kubectl CLI binary connection to kube API server running inside the box. Probably we need to just display the message
"Client binary install for kubernetes service is not support yet."

brgnepal:
@navidshaikh Fixed the message and added respective unit test.

test/vagrant-service-manager/installer_test.rb
```diff
@@ -71,6 +71,17 @@ module ServiceManager
...
         end
       end
+
+      describe 'Kubernetes' do
+        before do
+          @type = :kubernetes
+          @options = { box_version: 'adb' }
+        end
+
+        it 'should displays unsupported message' do
```
hferentschik: Nice that we can write tests now :-)

test/vagrant-service-manager/installer_test.rb
```diff
@@ -71,6 +71,17 @@ module ServiceManager
...
+          assert_raises(SystemExit) { @installer = Installer.new(@type, @machine, @machine.env, @options) }
```
hferentschik, line 82: But is this assertion really verifying the message? This is just a check for SystemExit, right?

locales/en.yml
```diff
@@ -186,6 +186,8 @@ en:
           Please visit access.redhat.com to download client binaries.
         service_not_enabled: |-
           '%{service}' service is not enabled.
+        kube_not_supportted: |-
+          Client binary installs for kubernetes service is not supported yet.
```
hferentschik: "Installation of Kubernetes client library via the install-cli command  is not supported yet"

If not, you need to at least adjust the current sentence. "installs" is plural, "is" singular. Does not match.

hferentschik:
I think we forgot to update the README in the main install-cli pull request. This is a good possibility to make up for it and also add the limitations of the command.

brgnepal:
@hferentschik Ya thanks for pointing out. Updating now.

test/vagrant-service-manager/installer_test.rb
budhrg: We can only verify exit at the moment. Don't know how to verify message.

test/vagrant-service-manager/installer_test.rb
brgnepal, line 82: We can only verify exit at the moment. Don't know how to verify message. 
Updated the test case description

brgnepal:
@hferentschik Updated with suggested changes.

README.adoc
```diff
@@ -150,10 +147,18 @@ only supported orchestrators are reported.
...
---
 The supported options for `service` are: Docker, OpenShift.
 
+1. `vagrant service-manager install-cli [service] [--cli-version] [--path]`
++
+Install the client binary for the specified `service`.
++
+The possible options are:
 
+  * `--cli-version`: Download the specified binary version.
```
hferentschik: "Download the binary in the specified version, if it exist"

What is the behavior if the version does not exist or wrong? What is the exit code? Does it throw an exception? What is the current behavior and what do we want it to do? I would think it returns a message of the form "Specified client binary version does not exist" and a non 0 return code. Is this what's happening?

README.adoc
```diff
@@ -150,10 +147,18 @@ only supported orchestrators are reported.
...
+  * `--path`: Download binary to the specified host path.
```
hferentschik: Download the binary to the specified path on the host

README.adoc
brgnepal: If the download url is inaccessible then following error message will be printed.

```
Download URL is not accessible.
Possible reason: Invalid version name
```

As, there is no simple way to validate the valid versions. One hard way is to somehow get all the released versions list and validate against that list.

hferentschik:
Superseded by pull request #317

PR #317, opened 2016-07-12
Fix 312 Error message when install-cli is used for Kubernetes

hferentschik:


hferentschik:
@budhrg Instead of trying to explain what I mean with "the test being to vague", I took the liberty to show you what I mean.

test/test_helper.rb
```diff
@@ -8,15 +8,26 @@
...
 
+class FakeUI
+  attr_reader :received_info_messages
+
+  def initialize
+    @received_info_messages = []
+  end
+
+  def info(*args)
+    # puts "#{args}"
```
brgnepal, line 26: Why this unwanted comment? Left out by mistake?

test/vagrant-service-manager/installer_test.rb
```diff
@@ -71,6 +73,24 @@ module ServiceManager
...
+        end
+
+        it 'should exit' do
+          begin
+            Installer.new(@type, @machine, @machine.env, @options)
+            refute(true, 'Installer should have exited')
+          rescue SystemExit => e
+            e.status.must_equal 126 # exited with failure status
+          end
+          @ui.received_info_messages.must_include 'Installation of Kubernetes client library via the install-cli ' \
```
brgnepal, line 90: :+1:

brgnepal:
Merged and thanks.

README.adoc
```diff
@@ -150,10 +147,18 @@ only supported orchestrators are reported.
...
+1. `vagrant service-manager install-cli [service] [--cli-version] [--path]`
++
+Install the client binary for the specified `service`.
++
+The possible options are:
 
+  * `--cli-version`: Download the binary in the specified version, if it exists.
+  * `--path`: Download the binary to the specified path on the host.
++
+The supported options for `service` are: Docker, OpenShift.
```
navidshaikh, line 161: I was wondering if we could have added a note about the default location as well, where the binaries will be downloaded.

brgnepal, line 161: That is added in the command output @navidshaikh

test/test_helper.rb
hferentschik, line 26: Because it would print into the default progress bar of the test run. I really like to keep this clean. On the other hand, when debugging it is sometimes handy to turn on. Hence, I left in in there commented out.

README.adoc
navidshaikh, line 161: @budhrg : I saw that, the message line after successful command execution prints the path where binary is downloaded, and it has to be there for completeness. I mentioned about docs since it makes sense to note down such kind of information in documentation as well.

brgnepal, line 161: @navidshaikh ACK

navidshaikh, line 161: Created the PR https://github.com/projectatomic/vagrant-service-manager/pull/324 with note about it, further discussions on an open PR! :-)

PR #328, opened 2016-07-14
Fix 309 Adding current provider to config.vm.box option in Vagrantfil…

### LINKED ISSUE #309: Unable to run Cucumber tests against ADB and CDK using libvirt
Using something like

```
$ bundle exec rake features BOX=cdk,adb PROVIDER=libvirt 
```

fails, whereas 

```
$ bundle exec rake features BOX=cdk,adb PROVIDER=virtualbox 
```

works.

The underlying problems is most likely somewhere in the underlying kvm/qemu which also caches virtual machines. The tests will alternate between adb and cdk boxes, but I think that on libvirt always the same box is used. We might need to clean kvm.qemu after each run.

hferentschik:
Fix #309 

…e used by Cucumber to avoid adb and cdk image to be cached under the same name in the libvirt pool

### DIFF
--- a/features/box-command.feature
+++ b/features/box-command.feature
@@ -14,7 +14,7 @@ Feature: Command output from box command
     end
 
     Vagrant.configure('2') do |config|
-      config.vm.box = '<box>'
+      config.vm.box = '<box>-<provider>'
       config.vm.box_url = 'file://../../.boxes/<box>-<provider>.box'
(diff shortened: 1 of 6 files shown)


hferentschik:
I changed temporarily the CI job config to run ADB and CDK. See if this works. Not sure, more of an experiment.

hferentschik:
Closing this pull request. This does still not cut it!?

PR #329, opened 2016-07-14
Fix #207: Removes SPEC file from repository and reference from README

### LINKED ISSUE #207: Remove vagrant-service-manager.spec and stop publishing gem as rpm
I am wondering why we publish as rpm. I find _vagrant-service-manager.spec_ confusing and I doubt/wonder what we gain by publishing an rpm. In the end it is just unnecessary overhead.

navidshaikh:
Fixes #207
  Since Ruby gem is the preferred method to install the plugin in system,
  this PR is removing the SPEC file from the git repository and also removes
  the reference for installing the plugin using RPM from README. If a user comes
  with usecase and request to provide the plugin packaged as RPM, we can add later.


hferentschik:
thanks

hferentschik:
Merged.

PR #334, opened 2016-07-14
Fix #313 Explicitly requiring rubyzip as dependency in gemspec. Despi…

### LINKED ISSUE #313: Plugin failed to load properly on Fedora
I installed vsm plugin on fedora 23 and it throws an error 

```
naina@dhcp35-1 ~]$ vagrant status 
Vagrant failed to initialize at a very early stage:
The plugins failed to load properly. The error message given is shown below.
cannot load such file -- zip
```

The same plugin version is working fine on windows , but for fedora it is throwing error while executing commands:
vagrant version,status,global-status,up etc.This issue is blocker for fedora
vsm version  = vagrant-service-manager (1.1.0.1)

hferentschik:
Fix #313  
…te being in the core Ruby packages, it seems this is still needed on Fedora

### DIFF
--- a/vagrant-service-manager.gemspec
+++ b/vagrant-service-manager.gemspec
@@ -16,4 +16,6 @@ Gem::Specification.new do |spec|
 
   spec.files         = `git ls-files -z`.split("\x0")
   spec.require_paths = ['lib']
+
+  spec.add_dependency 'rubyzip', '~> 1.2.0'
 end


hferentschik:
@LalatenduMohanty @budhrg can one of you guys tests this on Fedora? Pretty sure it works, but only quickly verified it within a Vagrant Fedora VM, so I could not fully start the ADB/CDK. However, I've seen the described error message and it went away. 

For the lazy https://dl.dropboxusercontent.com/u/459489/gems/vagrant-service-manager-1.3.0.dev.gem

hferentschik:
If it works we can cut a bug fix release (1.2.1)

vagrant-service-manager.gemspec
```diff
@@ -16,4 +16,6 @@ Gem::Specification.new do |spec|
 
   spec.files         = `git ls-files -z`.split("\x0")
   spec.require_paths = ['lib']
+
+  spec.add_dependency 'rubyzip'
```
brgnepal: Can you add version as `~> 1.2.0` as if there are systems with version `0.9` or less it will break again.
Refer link https://github.com/rubyzip/rubyzip/issues/90#issuecomment-23547717

LalatenduMohanty:
Tested on Fedora 23 and it works fine for me.

brgnepal:
@naina-verma found issue with Windows 10 vagrant version 1.8.4 as:

[issue comment](https://github.com/projectatomic/vagrant-service-manager/issues/313#issuecomment-232878702)

```
I tried the fix is working fine now on fedora .But, now its installation fails on windows with the following error 
https://gist.github.com/naina-verma/035ae882e339e5cfc73f89606251eabd

# NOTE - "vagrant-service-manager-1.2.0.libvirt.pre.fedora.pre.patch.gem" is build by me with exact as this PR with pinned to ~> 1.2.0 
```

 It is mostly related to local installation of gem which we are getting in Fedora 24 with vagrant 1.8.4 too.
This issue will be fixed in vagrant 1.8.5.

I verified this fix is working fine in Windows 10 vagrant 1.7.4(recommanded version)

LalatenduMohanty:
I can install the gem on OS X and `vagrant` is working fine with the plugin.

```
$ vagrant plugin install ~/Downloads/vagrant-service-manager-1.3.0.dev.gem 
Installing the '/Users/lmohanty/Downloads/vagrant-service-manager-1.3.0.dev.gem' plugin. This can take a few minutes...
Installed the plugin 'vagrant-service-manager (1.3.0.dev)'!
```

hferentschik:
> It is mostly related to local installation of gem which we are getting in Fedora 24 with vagrant 1.8.4 too.
> This issue will be fixed in vagrant 1.8.5.

@budhrg right, this is more likely the local gem install issue. Not related to what we do here. Note, on most Ruby's the dependency to rubyzip should not be needed. It is just Fedora's packaging which comes into play here.

vagrant-service-manager.gemspec
hferentschik: +1, updating pull request

brgnepal: @hferentschik could you update it?

brgnepal:
Merged and thanks.

hferentschik:
Cool. I have one more thing for 1.2.1 and then we cut the bug fix release ...

brgnepal:
@hferentschik this would be nice addition https://github.com/projectatomic/vagrant-service-manager/issues/341

LalatenduMohanty:
@budhrg are you suggesting #341 for version 1.2.1? if so why? IMO #341  is not a blocker at all for 1.2.1

brgnepal:
@LalatenduMohanty #341 is fixed with #342 and nice to have feature. Not a blocker.

LalatenduMohanty:
@budhrg what happens when this code change causes another blocker after release 1.2.1?

brgnepal:
hmmm @LalatenduMohanty why you feel so?

LalatenduMohanty:
@budhrg The idea is to minimize the risk. No code is 100% bug free and there is no guarantee that a code will work on all machines. All software goes through its life cycle i.e. new code has more bugs then we keep doing bug fixes and gradually it becomes stable enough. 
In this context we are trying to make 1.2.0 usable so our goal is to fix blocker bugs. Your suggestion is an enhancement. Hence it does not make sense to include an enhancement in a bug-fix release.

brgnepal:
@LalatenduMohanty ACK.

PR #340, opened 2016-07-15
Fix #339: Remove 1.3.0.dev from version.rb file

### LINKED ISSUE #339: Remove 1.3.0.dev version in version.rb file
I found `1.3.0.dev` in [version file](https://github.com/projectatomic/vagrant-service-manager/blob/master/lib/vagrant-service-manager/version.rb).

Is it by mistake?
If not then are we following any new policy to update version here? If yes then we need to discuss first.

brgnepal:
Fix #339


hferentschik:
1.3.0.dev is intentional to differentiate between the last stable release and the new dev version.

PR #342, opened 2016-07-15
Fix #341: Update gemspec with development dependencies

### LINKED ISSUE #341: Update gemspec with development dependencies
Updating gemspec with development dependencies will reflect the changes in rubygem's gem page.
This will help to quickly know the dependencies through rubygems instead of going through github repo and check Gemfile or gemspec.

Refer this nice blog post on [`roles of Gemfile and .gemspec` 2nd para](http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/)

Example gem:
screenshot from 2016-07-15 13-48-20

brgnepal:
Fix #341

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,19 +8,6 @@ group :development do
       ref: 'v1.8.4'
   gem 'vagrant-libvirt'              if RUBY_PLATFORM =~ /linux/i
   gem 'fog-libvirt', '0.0.3'         if RUBY_PLATFORM =~ /linux/i # https://github.com/pradels/vagrant-libvirt/issues/568
-  gem 'mechanize'
-  gem 'json'
-  gem 'cucumber', '~> 2.1'
-  gem 'aruba', '~> 0.13'
-  gem 'komenda', '~> 0.1.6'
-  gem 'launchy'
-  gem 'rake', '10.4.2'
-  gem 'rubocop'
-end
-
-group :test do
-  gem 'minitest'
-  gem 'mocha'
 end
 
 group :plugins do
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
-1 for release 1.2.1. Why are we even adding more things to 1.2.1. 1.2.1 is a release to fix the blocker which was caused by 1.2.0. What am I missing?

hferentschik:
I am torn on this one, since we are not able to move all development dependencies. the Vagrant, vagrant-libvirt and fog-libvirt will always have to stay in the Gemfile. I find that unfortunate. Also, we will always have to keep the plugin group. 

What we can do though is to "merge" the test group into the development one.

I also agree with @LalatenduMohanty, this is a bug fix release. We should not change more code than necessary. I'd rather just fix the rubyzip thing. The other issues we added to 1.2.1 are just doc changes.

brgnepal:
@LalatenduMohanty @hferentschik Ok we will add in next major version then.

hferentschik:
> Ok we will add in next major version then.

-1 from my side. Why would you split up development dependencies in two places. That is confusing imo. Sure, if we could move all of them, it would make sense, but gemsepc does not allow for things like github references or the conditional inclusion of dependencies.

Also from a user perspective we would be displaying incomplete information. The most important dependency would be missing - Vagrant.

I fully agree it would be desirable to have all dependencies in gemspec, but given this is not possible, I'd opt for keeping them combined in one place.

brgnepal:
Closing as we are not having consensus on adding dependencies in `.gemspec`

PR #349, opened 2016-07-21
Fix #348, #357: Handle 302 redirection in install-cli URL validation

### LINKED ISSUE #348: 'install-cli' not working for Openshift
Hi,

Downloading client binary for openshift seems to be not working. It is showing 'Download URL not accessible error'.

```
➜  vagrant-service-manager git:(vsm-1.2.1) ✗ vagrant version
Installed Version: 1.8.4

# Binary now available at /home/budhram/.vagrant.d/data/service-manager/bin/docker/1.9.1/docker
# run binary as:
# docker <command>
export PATH=/home/budhram/.vagrant.d/data/service-manager/bin/docker/1.9.1:$PATH

# eval "$(VAGRANT_NO_COLOR=1 vagrant service-manager install-cli docker | tr -d '\r')"

$ vagrant service-manager install-cli openshift
Download URL is not accessible.
Possible reason: Invalid version name
```

### LINKED ISSUE #357: Unable to download cli if user used latest tag or a alpha/beta tag.
**OS details:**
Linux: Fedora24

**Provider:**
VirtualBox

**Output of `vagrant -v`:**

```
N/A
```

```
N/A
```

**Output of `vagrant service-manager box version`:**

```
Atomic Developer Bundle (ADB) 2.2.2
```

**Steps to reproduce the issue:**
1. Put **latest** tag in the Vagrantfile
2. vagrant up
3. vagrant service-manager install-cli openshift

**Describe the results you received:**

```
bad URI(is not URI?): https://github.com/openshift/origin/releases/tag/v1.3.0
.2
250
61
05
```

**Describe the results you expected:**

```
A better user warning if specific tag is not available to github.
```

brgnepal:
Fix #357 , #348 
- Added unit test to test URL validation

### DIFF
--- a/.config/cucumber.yml
+++ b/.config/cucumber.yml
@@ -1,7 +1,7 @@
 # config/cucumber.yml
 ##YAML Template
 ---
-default: --profile html
+default: --profile html --tags ~@ci
 
 pretty: --format pretty -b
 html: --format progress --format html --out=build/features_report.html -b
--- a/features/adb-openshift.feature
+++ b/features/adb-openshift.feature
@@ -1,7 +1,6 @@
-Feature: Command output from various OpenShift related commands
-  service-manager should return the correct output from commands affecting OpenShift
+Feature: Command output from various OpenShift related commands in ADB
+  service-manager should return the correct output from commands affecting OpenShift in ADB
 
-  @openshift
   Scenario Outline: Boot and execute commands
     Given box is <box>
     And provider is <provider>
@@ -18,11 +17,11 @@ Feature: Command output from various OpenShift related commands
       config.vm.box_url = 'file://../../.boxes/<box>-<provider>.box'
       config.vm.network :private_network, ip: '<ip>'
       config.vm.synced_folder '.', '/vagrant', disabled: true
-      config.servicemanager.services = 'docker'
-      config.vm.provision "shell", inline: <<-SHELL
-        systemctl enable openshift 2>&1
-        systemctl start openshift | true
-      SHELL
+      config.servicemanager.services = 'docker, openshift'
+
+      config.servicemanager.openshift_image_tag = 'v1.2.1'
+      config.servicemanager.openshift_image_name = 'openshift/origin'
+      config.servicemanager.openshift_docker_registry = 'docker.io'
     end
     """
 
@@ -36,7 +35,7 @@ Feature: Command output from various OpenShift related commands
     # To use OpenShift CLI, run: oc login https://<ip>:8443
     export OPENSHIFT_URL=https://<ip>:8443
     export OPENSHIFT_WEB_CONSOLE=https://<ip>:8443/console
-    export DOCKER_REGISTRY=hub.openshift.rhel-cdk.<ip>.xip.io
+    export DOCKER_REGISTRY=hub.openshift.centos7-adb.<ip>.xip.io
 
     # run following command to configure your shell:
     # eval "$(vagrant service-manager env openshift)"
@@ -48,14 +47,14 @@ Feature: Command output from various OpenShift related commands
     """
     OPENSHIFT_URL=https://<ip>:8443
     OPENSHIFT_WEB_CONSOLE=https://<ip>:8443/console
-    DOCKER_REGISTRY=hub.openshift.rhel-cdk.<ip>.xip.io
+    DOCKER_REGISTRY=hub.openshift.centos7-adb.<ip>.xip.io
     """
 
     When I successfully run `bundle exec vagrant service-manager env`
     Then stdout from "bundle exec vagrant service-manager env" should contain "export DOCKER_HOST=tcp://<ip>:2376"
     And stdout from "bundle exec vagrant service-manager env" should match /export DOCKER_CERT_PATH=.*\/.vagrant\/machines\/cdk\/virtualbox\/docker/
     And stdout from "bundle exec vagrant service-manager env" should contain "export DOCKER_TLS_VERIFY=1"
-    And stdout from "bundle exec vagrant service-manager env" should contain "export DOCKER_API_VERSION=1.21"
+    And stdout from "bundle exec vagrant service-manager env" should contain "export DOCKER_API_VERSION=1.2\d"
     And stdout from "bundle exec vagrant service-manager env" should match /# eval "\$\(vagrant service-manager env\)"/
     And stdout from "bundle exec vagrant service-manager env" should contain:
     """
@@ -64,13 +63,18 @@ Feature: Command output from various OpenShift related commands
     # To use OpenShift CLI, run: oc login https://<ip>:8443
     export OPENSHIFT_URL=https://<ip>:8443
     export OPENSHIFT_WEB_CONSOLE=https://<ip>:8443/console
-    export DOCKER_REGISTRY=hub.openshift.rhel-cdk.<ip>.xip.io
+    export DOCKER_REGISTRY=hub.openshift.centos7-adb.<ip>.xip.io
 
     # run following command to configure your shell:
     # eval "$(vagrant service-manager env)"
     """
 
+    When I successfully run `bundle exec vagrant up --provider <provider>`
+    And I run `bundle exec vagrant service-manager install-cli openshift`
+    Then the exit status should be 0
+    And the binary "oc" should be installed
+
     Examples:
       | box   | provider   | ip          |
-      | cdk   | virtualbox | 10.10.10.42 |
-      | cdk   | libvirt    | 10.10.10.42 |
+      | adb   | virtualbox | 10.10.10.42 |
+      | adb   | libvirt    | 10.10.10.42 |
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -5,9 +5,8 @@ require 'find'
 ###############################################################################
 # Aruba config and Cucumber hooks
 ###############################################################################
-
 Aruba.configure do |config|
-  config.exit_timeout = 300
+  config.exit_timeout = 3600
   config.activate_announcer_on_command_failure = [:stdout, :stderr]
   config.working_directory = 'build/aruba'
 end
@@ -19,7 +18,7 @@ end
 
 After do |_scenario|
   if File.exist?(File.join(aruba.config.working_directory, 'Vagrantfile'))
-    Komenda.run('bundle exec vagrant destroy -f', cwd: aruba.config.working_directory, fail_on_fail: true)
+    #Komenda.run('bundle exec vagrant destroy -f', cwd: aruba.config.working_directory, fail_on_fail: true)
     if ENV.key?('CUCUMBER_RUN_PROVIDER')
       # if we have more than one provider we need to wait between scenarios in order to allow for
       # proper cleanup/shutdown of virtualization framework
@@ -28,7 +27,7 @@ After do |_scenario|
   end
 
   # Remove the created Vagrant home dir
-  FileUtils.rmtree(ENV['VAGRANT_HOME']) if File.directory? ENV['VAGRANT_HOME']
+  # FileUtils.rmtree(ENV['VAGRANT_HOME']) if File.directory? ENV['VAGRANT_HOME']
 end
 
 ###############################################################################
--- a/lib/vagrant-service-manager/binary_handlers/binary_handler.rb
+++ b/lib/vagrant-service-manager/binary_handlers/binary_handler.rb
@@ -13,7 +13,7 @@ module VagrantPlugins
       }.freeze
       VERSION_CMD = {
         docker: "docker version --format '{{.Server.Version}}'",
-        openshift: "oc version | grep 'oc' | grep -oE '[0-9.]+'"
+        openshift: "oc version | grep -oE 'oc v([0-9a-z.]+-?[a-z0-9]*.?[0-9])' | sed -E 's/oc v//'"
       }.freeze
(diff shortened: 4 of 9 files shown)


navidshaikh:
@budhrg : While testing this, I found https://github.com/projectatomic/vagrant-service-manager/issues/354

brgnepal:
@navidshaikh Is this good to go ?

hferentschik:
I'll test it as well

budhrg:
@hferentschik Why removed milestone? I think we should release this bug fix as soon as possible.

test/vagrant-service-manager/binary_handlers/adb_docker_binary_handler_test.rb
```diff
@@ -54,6 +54,11 @@ module ServiceManager
       end
 
       it 'should validate download url' do
+        @handler.build_download_url
+        assert_silent { @handler.validate_url }
```
hferentschik: This is an odd test. You are saying that just because there is no output, the validation succeeded. Why does validate_url not return something _true_ or something? Can you not test some actual URLs and also trigger the exceptional case. You might want to pass the url into the method instead of using @url. This way you can test this method much easier.

lib/vagrant-service-manager/binary_handlers/binary_handler.rb
```diff
@@ -100,7 +100,7 @@ def validate_url
         req.use_ssl = true if url.scheme == 'https'
         res = req.request_head(url.path)
 
-        unless res.code == '200'
+        unless %w(200 302).include? res.code
```
hferentschik, line 103: So what is going on there? Is there ourdays a redirection in the oc download URL? I am pretty sure this was not the case before, right?

brgnepal, line 103: For openshift url, status code seems to be 302(redirection).

test/vagrant-service-manager/binary_handlers/adb_docker_binary_handler_test.rb
brgnepal: > pass the url into the method instead of using @url.

What will be method signature here ? `def validate_url(url)` ? This doesn't seems to be object oriented.

> Why does validate_url not return something true or something?

Thought of that but then we are capturing the return value in `install` method. Whats its purpose of returning?

brgnepal: > assert_silent { @handler.validate_url }

@hferentschik Here, I am testing whether the download URL built from already provided options in `before` block is fine enough and should not throw any kind of exception against what it is expected to have some exception on invalid URL which is covered by the test case just below this test case.

hferentschik: > What will be method signature here ? def validate_url(url) ? This doesn't seems to be object oriented.

It is object oriented and increased test-ability atm. 

> Thought of that but then we are capturing the return value in install method. Whats its purpose of returning?

A more useful method and test-ability

> is fine enough and should not throw any kind of exception against what it is expected to have some exception on invalid URL which is covered by the test case just below this test case.

Sure, but what you are asserting is that the command is silent. That's about it. I guess what you are after is _assert_nothing_raised_. And if you use this assertion, you really want to assert the exceptional case as well. 

That said, why don't you add a Cucumber test for the oc cli install? This will once and for all show whether this works end to end.

brgnepal: Didn't know about `assert_nothing_raise`. I was expecting to do it but found `assert_silent` somewhere while googling.

navidshaikh:
I am getting bad UI error

```
➜  vagrant-service-manager git:(fix-348) ✗ bundle exec vagrant service-manager status    
Configured services:
docker - running
openshift - running
kubernetes - stopped

➜  vagrant-service-manager git:(fix-348) ✗ bundle exec vagrant service-manager box version
Atomic Developer Bundle (ADB) 2.2.0

➜  vagrant-service-manager git:(fix-348) ✗ bundle exec vagrant service-manager install-cli openshift 

bad URI(is not URI?): https://github.com/openshift/origin/releases/tag/v1.3.0
.2
250
61
05

➜  vagrant-service-manager git:(fix-348) ✗ echo $?
126
```

brgnepal:
@navidshaikh I noticed that you are using version `Atomic Developer Bundle (ADB) 2.2.0` which is 2 version old than current `ADB 2.2.2`.

Could you verify the same with `ADB 2.2.2`?

Also not sure how you are getting `1.3.0` tag url since origin latest is `1.2.1`. https://github.com/openshift/origin/tags

lib/vagrant-service-manager/binary_handlers/binary_handler.rb
hferentschik, line 103: Again, that's why we need a verified end to end test. TBH, I am surprised. I thought I tested it, but maybe it was only 'docker' after all.

brgnepal, line 103: Ya, little weird. For docker response code is 200.

navidshaikh:
@budhrg : Even after updating the adb to 2.2.2, I am getting same error. @praveenkumar also filed same issue https://github.com/projectatomic/vagrant-service-manager/issues/357

brgnepal:
@navidshaikh I would like to debug more on it. Could you please provide me the logs with `--debug` flag to get start with debug?

hferentschik:
What's the status here? Can we get a working _oc_ download soon?

brgnepal:
@hferentschik I updated the unit tests.

Will add cucumber test as part of CDK support implementation for install-cli.

features/adb-openshift.feature
```diff
@@ -0,0 +1,80 @@
+Feature: Command output from various OpenShift related commands in ADB
+  service-manager should return the correct output from commands affecting OpenShift in ADB
+
+  @ci
```
brgnepal: To be run only through ci pipeline

features/adb-openshift.feature
```diff
@@ -0,0 +1,80 @@
...
+    # To use OpenShift CLI, run: oc login https://<ip>:8443
+    export OPENSHIFT_URL=https://<ip>:8443
+    export OPENSHIFT_WEB_CONSOLE=https://<ip>:8443/console
+    export DOCKER_REGISTRY=hub.openshift.rhel-cdk.<ip>.xip.io
+
+    # run following command to configure your shell:
+    # eval "$(vagrant service-manager env)"
+    """
+
+    When I run `bundle exec vagrant service-manager install-cli openshift`
```
brgnepal: @hferentschik  Install-cli openshift acceptance test

features/support/env.rb
```diff
@@ -5,6 +5,9 @@
 ###############################################################################
 # Aruba config and Cucumber hooks
 ###############################################################################
+BINARY_MAP = {
```
brgnepal: Needed as openshift has binary name as 'oc'

brgnepal:
retest this please

hferentschik:
@budhrg The oc install seems to fail on CI

.config/cucumber.yml
```diff
@@ -1,7 +1,7 @@
 # config/cucumber.yml
 ##YAML Template
 ---
-default: --profile html
+default: --profile html --tags ~@ci
```
hferentschik, line 4: +1, except I'd call _ci_ either _ci-only_ or maybe we should even call it _long_running_. The latter makes also clear why we don't want to run this per default.

I'd also add a _all_ profile, so that we can run easily  on the dev machine all tests if we want to

features/adb-openshift.feature
hferentschik: +1, just see comment about tag name above.

features/adb-openshift.feature
```diff
@@ -0,0 +1,80 @@
...
+
+    Vagrant.configure('2') do |config|
+      config.vm.box = '<box>'
+      config.vm.box_url = 'file://../../.boxes/<box>-<provider>.box'
+      config.vm.network :private_network, ip: '<ip>'
+      config.vm.synced_folder '.', '/vagrant', disabled: true
+      config.servicemanager.services = 'docker'
+      config.vm.provision "shell", inline: <<-SHELL
+        systemctl enable openshift 2>&1
+        systemctl start openshift | true
```
hferentschik: OOI, why the _| true_? This means the provisioning will always be "successful", even is starting the service failed, no?

brgnepal: Ya forgot to remove it. Just copied it from openshift.feature. will remove it

features/adb-openshift.feature
```diff
@@ -0,0 +1,80 @@
...
+    export OPENSHIFT_WEB_CONSOLE=https://<ip>:8443/console
+    export DOCKER_REGISTRY=hub.openshift.rhel-cdk.<ip>.xip.io
+
+    # run following command to configure your shell:
+    # eval "$(vagrant service-manager env)"
+    """
+
+    When I run `bundle exec vagrant service-manager install-cli openshift`
+    Then the exit status should be 0
+    And the binary "openshift" should be installed
```
hferentschik: `And the binary "oc" should be installed` - The step definition is for testing an installed binary. The service for which you install the binary is _openshift_, but you need to test for the _oc_ binary.

features/support/env.rb
hferentschik: I don't think you need that. Not if you, as intended,  use the actual binary name in the step

test/vagrant-service-manager/binary_handlers/adb_docker_binary_handler_test.rb
```diff
@@ -54,6 +54,11 @@ module ServiceManager
       end
 
       it 'should validate download url' do
+        @handler.build_download_url
+        @handler.validate_url.must_equal true
```
hferentschik, line 58: +1

features/support/env.rb
brgnepal: Got you :smile:

hferentschik:
@budhrg Let's add something to the README in the acceptance test section on how to use profiles. Something like:

There is a default profile which does not run long running tests. To exclude a test from running per default you annotate it with _ci-only_/_long_running_ (whatever tag we choose in the end). Also to run all tests there is (will be) an _all_ profile. Profiles can be activated via the _CUCUMBER_OPTS_ environment variable:

```
bundle exec rake features CUCUMBER_OPTS='-p all' PROVIDER=libvirt BOX=adb,cdk
```

Last but not least, I think we can remove all the other tag we have so far. As said, I added them thinking that's the way to run a single feature file, but I think they are obsolete. WDYT?

hferentschik:
@budhrg Running test locally to see if it works on my machine

hferentschik:
@budhrg I think we should increase the Aruba timeout as well. Currently it is 5 min (300 s). See _env.rb_. On my first attempt to run this test, the _vagrant up_ got aborted, because it took longer than 5 minutes to provision Openshift on ADB.

hferentschik:
@budhrg I see the same error on my machine as on CI. This is even before the install-cli part gets executed. 

If you add to _cucumber.yml_:

```
all: --profile html
```

and then:

```
bundle exec rake features CUCUMBER_OPTS='-p all' FEATURE=features/adb-openshift.feature 
```

Are you saying the this works on your machine? For all I can tell, the OpenShift service does not start properly.

lib/vagrant-service-manager/binary_handlers/binary_handler.rb
```diff
@@ -13,7 +13,7 @@ class BinaryHandler
       }.freeze
       VERSION_CMD = {
         docker: "docker version --format '{{.Server.Version}}'",
-        openshift: "oc version | grep 'oc' | grep -oE '[0-9.]+'"
+        openshift: "oc version | grep -oE 'oc v([0-9a-z.]+-?[a-z0-9]*.?[0-9])' | sed -E 's/oc v//'"
```
brgnepal, line 16: Fix "#357"

hferentschik, line 16: :-) Told you the end to end test will shake things out :-)

features/support/env.rb
```diff
@@ -19,7 +18,7 @@
 
 After do |_scenario|
   if File.exist?(File.join(aruba.config.working_directory, 'Vagrantfile'))
-    Komenda.run('bundle exec vagrant destroy -f', cwd: aruba.config.working_directory, fail_on_fail: true)
+    #Komenda.run('bundle exec vagrant destroy -f', cwd: aruba.config.working_directory, fail_on_fail: true)
```
hferentschik, line 21: watch out here. You commented out some test harness setup/teardown stuff and accidentally committed it.

features/support/env.rb
```diff
@@ -28,7 +27,7 @@
   end
 
   # Remove the created Vagrant home dir
-  FileUtils.rmtree(ENV['VAGRANT_HOME']) if File.directory? ENV['VAGRANT_HOME']
+  # FileUtils.rmtree(ENV['VAGRANT_HOME']) if File.directory? ENV['VAGRANT_HOME']
```
hferentschik, line 30: Same as above.

hferentschik:
@budhrg I created pull request #363 as follow up, cleaning up some of the remaining details we discussed as well as adding some docs to the README. Let me know what you think.

hferentschik:
Superseded by pull request #363

PR #368, opened 2016-08-04
Fix #360: install-cli for docker and openshift in CDK

### LINKED ISSUE #360: Support install-cli for docker and openshift in CDK
Currently, `install-cli` command is only supported in ADB for 'docker' and 'openshift' services.
Add support of it for CDK box too.

brgnepal:
Fix #360

### DIFF
--- a/features/cdk-openshift.feature
+++ b/features/cdk-openshift.feature
@@ -18,11 +18,7 @@ Feature: Command output from various OpenShift related commands in CDK
       config.vm.box_url = 'file://../../.boxes/<box>-<provider>.box'
       config.vm.network :private_network, ip: '<ip>'
       config.vm.synced_folder '.', '/vagrant', disabled: true
-      config.servicemanager.services = 'docker'
-      config.vm.provision "shell", inline: <<-SHELL
-        systemctl enable openshift 2>&1
-        systemctl start openshift | true
-      SHELL
+      config.servicemanager.services = 'docker, openshift'
     end
     """
 
@@ -69,6 +65,10 @@ Feature: Command output from various OpenShift related commands in CDK
     # run following command to configure your shell:
     # eval "$(vagrant service-manager env)"
     """
+
+    When I run `bundle exec vagrant service-manager install-cli openshift`
+    Then the exit status should be 0
+    And the binary "oc" should be installed
 
     Examples:
       | box   | provider   | ip          |
--- a/lib/vagrant-service-manager/binary_handlers/adb_binary_handler.rb
+++ b/lib/vagrant-service-manager/binary_handlers/adb_binary_handler.rb
@@ -4,52 +4,6 @@ module VagrantPlugins
       def initialize(machine, env, options)
         super(machine, env, options)
       end
-
-      def build_archive_path
-        @archive_file_path = "#{@temp_bin_dir}/#{File.basename(@url)}"
-      end
-
-      def ensure_binary_and_temp_directories
-        FileUtils.mkdir_p(bin_dir) unless File.directory?(bin_dir)
-        FileUtils.mkdir_p(@temp_bin_dir) unless File.directory?(@temp_bin_dir)
-      end
-
-      def download_archive
-        return @skip_download = true if File.exist?(@archive_file_path)
-        Vagrant::Util::Downloader.new(@url, @archive_file_path).download!
-      rescue Vagrant::Errors::DownloaderError => e
-        @ui.error e.message
-        exit 126
-      end
-
-      def prepare_binary
-        tmp_binary_file_path = @archive_file_path
-
-        # If binary is in archive format, extract it
-        if binary_archived?
-          tmp_binary_file_path = "#{archive_dir_name}/#{binary_name}"
-          archive_handler_class.new(@archive_file_path, tmp_binary_file_path, file_regex).unpack
-        end
-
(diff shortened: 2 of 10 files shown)


features/cdk-openshift.feature
```diff
@@ -18,11 +18,7 @@ Feature: Command output from various OpenShift related commands in CDK
       config.vm.box_url = 'file://../../.boxes/<box>-<provider>.box'
       config.vm.network :private_network, ip: '<ip>'
       config.vm.synced_folder '.', '/vagrant', disabled: true
-      config.servicemanager.services = 'docker'
-      config.vm.provision "shell", inline: <<-SHELL
-        systemctl enable openshift 2>&1
-        systemctl start openshift | true
-      SHELL
+      config.servicemanager.services = 'docker, openshift'
```
hferentschik, line 21: nice cleanup

features/cdk-openshift.feature
```diff
@@ -70,6 +66,10 @@ Feature: Command output from various OpenShift related commands in CDK
     # eval "$(vagrant service-manager env)"
     """
 
+    When I run `bundle exec vagrant service-manager install-cli openshift`
+    Then the exit status should be 0
+    And the binary "oc" should be installed
+
```
hferentschik, line 72: nice

lib/vagrant-service-manager/binary_handlers/adb_binary_handler.rb
```diff
@@ -4,52 +4,6 @@ class ADBBinaryHandler < BinaryHandler
       def initialize(machine, env, options)
         super(machine, env, options)
       end
-
```
hferentschik, line 7: I am a bit surprised that you pull that in binary handler, especially since you rightfully extend the CDK handlers from the ADB ones.

lib/vagrant-service-manager/binary_handlers/cdk_docker_binary_handler.rb
```diff
@@ -1,6 +1,7 @@
 module VagrantPlugins
   module ServiceManager
-    class CDKDockerBinaryHandler < CDKBinaryHandler
+    # Currently client binary installation of docker for CDK is same as ADB
+    class CDKDockerBinaryHandler < ADBDockerBinaryHandler
```
hferentschik, line 4: +1

lib/vagrant-service-manager/binary_handlers/cdk_openshift_binary_handler.rb
```diff
@@ -1,7 +1,13 @@
 module VagrantPlugins
   module ServiceManager
-    class CDKOpenshiftBinaryHandler < CDKBinaryHandler
+    # Currently client binary installation of openshift for CDK is same as ADB
+    class CDKOpenshiftBinaryHandler < ADBOpenshiftBinaryHandler
+      # Default to latest stable origin oc version for CDK as it is different than
+      # OSE oc version running inside CDK development environment
+      LATEST_OC_VERSION = '1.2.1'.freeze
```
hferentschik, line 7: Does that mean that whenever a new OSE comes along we need to new service-manager release as well?

lib/vagrant-service-manager/binary_handlers/cdk_openshift_binary_handler.rb
```diff
@@ -1,7 +1,13 @@
...
+
       def initialize(machine, env, options)
+        options['--cli-version'] = LATEST_OC_VERSION unless options['--cli-version']
```
hferentschik, line 10: Except right now we don't have a --cli-version, exposed for the user, right? I believe though we have an issue for this.

lib/vagrant-service-manager/binary_handlers/cdk_openshift_binary_handler.rb
LalatenduMohanty, line 7: I think the major versions should match though like OSE 3.2 should use origin 1.2 OC client . Similarly OSE 3.3 should use Origin 1.3 OC client.  But we should avoid VSM release for OSE releases

lib/vagrant-service-manager/binary_handlers/cdk_openshift_binary_handler.rb
brgnepal, line 10: @hferentschik  `--cli-version` option is still supported for CDK client binary installation. Only in case of openshift, since there was no way to get the origin compatible version from CDK, I have pinned it to latest stable release of oc i.e `1.2.1`.

lib/vagrant-service-manager/binary_handlers/cdk_openshift_binary_handler.rb
brgnepal, line 7: > Does that mean that whenever a new OSE comes along we need to new service-manager release as well?

@hferentschik  No as @LalatenduMohanty  said most of latest client from oc should work with latest OSE. 
Now, to update this version, we can decide after how many iterations we should update.

Anyway, user always has a way to install the latest binary with `--cli-version`.

lib/vagrant-service-manager/binary_handlers/adb_binary_handler.rb
brgnepal, line 7: @hferentschik Ya but I believe most of the higher level operations like download archive, extract it, binary path etc are almost same for all handlers except `build_download_url` which may be different for every handler.

PR #373, opened 2016-08-11
Fix #372: Detect already downloaded oc binary in Windows

### LINKED ISSUE #372: Detect the already downloaded oc binary in Windows
Hi,

If user has 'oc' binary already available in machine which get downloaded when installing [Devsuite](https://github.com/redhat-developer-tooling/developer-platform-install) then detect it from service-manager plugin and display this path during `install-cli` command instead of fresh download.

[Devsuite](https://github.com/redhat-developer-tooling/developer-platform-install) unpacks OpenShift Origin CLI to `${target.location}/cdk/bin` where `target.location` can be any drive letter which user choose while installing devsuite. Devsuite also add this path to `PATH` env variable.

brgnepal:
Fix #372


naina-verma:
@budhrg , this vagrant-service-manager gem is  tested on windows 10. its working fine.
It shows up path of existing binary as a result .

PR #385, opened 2016-08-30
Fix #384: don't execute os_variant for unsupported box

### LINKED ISSUE #384: "Vagrant attempted to execute the capability 'os_variant'" issue happening again
This is related to https://github.com/projectatomic/vagrant-service-manager/issues/220

Looks like https://github.com/projectatomic/vagrant-service-manager/commit/e5f4d4d0e030bbd12d513746e1034489b9c86b1c removed the rescue in lib/vagrant-service-manager/service.rb that previously fixed this error.

Relevant line: https://github.com/projectatomic/vagrant-service-manager/commit/e5f4d4d0e030bbd12d513746e1034489b9c86b1c#diff-a81916b9cc51834172e560725e66491cL34

**OS details:**
CoreOS Beta 899.7.0

**Provider:**
VirtualBox

**Output of `vagrant -v`:**

```
Vagrant 1.8.1
```

```
vagrant-hostmanager (1.8.5)
vagrant-ohai (0.1.13)
vagrant-omnibus (1.4.1)
vagrant-registration (1.2.3)
vagrant-service-manager (1.3.1)
vagrant-share (1.1.5, system)
vagrant-triggers (0.5.3)
```

brgnepal:
Fix #384 

Fixed the regression due to https://github.com/projectatomic/vagrant-service-manager/commit/e5f4d4d0e030bbd12d513746e1034489b9c86b1c

### DIFF
--- a/lib/vagrant-service-manager/service.rb
+++ b/lib/vagrant-service-manager/service.rb
@@ -16,8 +16,13 @@ module VagrantPlugins
 
       def call(env)
         @app.call(env)
-        return false unless SUPPORTED_BOXES.include? @machine.guest.capability(:os_variant)
-        @service_hooks.each(&:execute)
+
+        if SUPPORTED_BOXES.include? @machine.guest.capability(:os_variant)
+          return false unless SUPPORTED_BOXES.include? @machine.guest.capability(:os_variant)
+          @service_hooks.each(&:execute)
+        end
+      rescue Vagrant::Errors::GuestCapabilityNotFound => e
+        PluginLogger.debug e.message
       end
 
       private


LalatenduMohanty:
@budhrg we should put some tests for this PR, may be some unit tests. Otherwise LGTM

lib/vagrant-service-manager/service.rb
```diff
@@ -16,8 +16,13 @@ def initialize(app, env)
...
         @app.call(env)
-        return false unless SUPPORTED_BOXES.include? @machine.guest.capability(:os_variant)
-        @service_hooks.each(&:execute)
+
+        if SUPPORTED_BOXES.include? @machine.guest.capability(:os_variant)
+          return false unless SUPPORTED_BOXES.include? @machine.guest.capability(:os_variant)
+          @service_hooks.each(&:execute)
+        end
+      rescue Vagrant::Errors::GuestCapabilityNotFound
+        # Do nothing if supported box variant not found
```
navidshaikh: Can we log the error?

brgnepal: @navidshaikh Probably, we don't want to do that who is using different box and have nothing to do with adb/cdk box. It would be unnecessary thing for the user.
And, if we do that user might come back complaining that vsm is throwing this error etc.

We might add debug logs here which might be visible during `--debug` option only but again is it necessary since its not relevant for us if box is not adb/cdk altogether.

LalatenduMohanty: +1 for debug log as users who are not using ADB/CDK would not bother about VSM. But debug log will help in case user is trying to make VSM work with anything other than ADB/CDK.

brgnepal:
@LalatenduMohanty @navidshaikh : Done with logging error as debug log which is only visible during `--debug` option.

brgnepal:
retest this please

brgnepal:
retest this please

naina-verma:
@budhrg ,LGTM!

PR #391, opened 2016-09-27
Acceptance test for --cli-version and --path options and updated help for install-cli

### LINKED ISSUE #326: Implement the --cli-version and --path options for install-cli
These options are not properly implemented yet. The help does not mention them, but they are mentioned in the README (going to remove it together with the Cucumber test which does not seem to do anything).

brgnepal:
Fix #326


brgnepal:
@hferentschik , any idea how to test `--path` option?

Suppose, if I run command `vagrant service-manager install-cli docker --path <path>`

How do I validate this `path` in terms of current acceptance test framework?

hferentschik:
@budhrg let me have a look into it, starting with CI ...

hferentschik:
@budhrg, could you please push a rebased version of this pull request? This should also trigger the hopefully working CI job again.

brgnepal:
@hferentschik It looks like rubocop check is not _really_ working locally

```
➜  vagrant-service-manager git:(fix-326) ✗ bundle exec rake rubocop
Running RuboCop...
Inspecting 47 files
...............................................

47 files inspected, no offenses detected
```

brgnepal:
@hferentschik Pushing rebased version will not work. Some "guard related warning" in new changes I committed.
Will fix that and push along with rebase.

brgnepal:
@hferentschik  Rubocop version was the culprit :wink: 

```
➜  vagrant-service-manager git:(fix-326) ✗ rubocop -v
0.42.0
➜  vagrant-service-manager git:(fix-326) ✗ rubocop lib/vagrant-service-manager/installer.rb
Inspecting 1 file
.

1 file inspected, no offenses detected
➜  vagrant-service-manager git:(fix-326) ✗ gem update rubocop
Updating installed gems
Updating rubocop
Fetching: rubocop-0.43.0.gem (100%)
Successfully installed rubocop-0.43.0
Parsing documentation for rubocop-0.43.0
Installing ri documentation for rubocop-0.43.0
Installing darkfish documentation for rubocop-0.43.0
Done installing documentation for rubocop after 12 seconds
Parsing documentation for rubocop-0.43.0
Done installing documentation for rubocop after 6 seconds
Gems updated: rubocop
➜  vagrant-service-manager git:(fix-326) ✗ rubocop lib/vagrant-service-manager/installer.rb
Inspecting 1 file
C

Offenses:

lib/vagrant-service-manager/installer.rb:39:9: C: Use a guard clause instead of wrapping the code inside a conditional expression.
        if @options.key?('--path') && !File.exist?(File.dirname(@options['--path']))
        ^^

1 file inspected, 1 offense detected
```

brgnepal:
retest this please

brgnepal:
retest this please

brgnepal:
@hferentschik Build passing now.

I am unable to add the test for `--path` option. How you can put path and verify it in aruba test framework?

hferentschik:
> I am unable to add the test for --path option. How you can put path and verify it in aruba test framework?

What exactly is the problem?

Gemfile
```diff
@@ -13,7 +13,7 @@ group :development do
   gem 'komenda', '~> 0.1.6'
   gem 'launchy'
   gem 'rake', '10.4.2'
-  gem 'rubocop'
+  gem 'rubocop', '~> 0.43.0'
```
hferentschik: Can you explain this change? How come that when I started from scratch as well, my Rubocop run worked, whereas on CI I got the errors? Are you saying different versions are getting installed depending on the environment?

brgnepal:
> What exactly is the problem?

What path to specify in command when I say

```
When I run `bundle exec vagrant service-manager install-cli docker --path <which path>`
And the binary "docker" should be installed with path "which path" (how do I verify here?)
```

lib/vagrant-service-manager/installer.rb
```diff
@@ -5,15 +5,15 @@ def initialize(type, machine, env, options)
         @type = type
         @machine = machine
         @env = env
-        @box_version = options[:box_version]
+        @options = { type: @type }.merge(options)
```
hferentschik: Is this not a bit odd? You pass _type_ and _option_ to this method, but then add _@type_ to the options as well? Would it not make more sense to add the type of the options before and just pass the options parameter?

hferentschik: Alternatively, if _options_ is really just the options passed to the CLI and _type_ something you determine, keep passing both parameters along.

test/vagrant-service-manager/installer_test.rb
```diff
@@ -29,7 +29,7 @@ module ServiceManager
           @installer.instance_variable_get('@type').must_equal @type
           @installer.instance_variable_get('@machine').must_equal @machine
           @installer.instance_variable_get('@env').must_equal @machine.env
-          @installer.instance_variable_get('@box_version').must_equal 'adb'
+          @installer.instance_variable_get('@options').must_equal({ type: @type }.merge(@options))
```
hferentschik: Even the test seems odd with this merge thing you are doing. Btw, I don't think these tests really do much. They are pretty much self fulfilling. I'd go so far as to say let's remove them

features/install-cli.feature
```diff
@@ -52,6 +54,10 @@ Feature: Command behavior of client side tools installation
     Then the exit status should be 0
     And the binary "docker" should be installed
 
+    When I run `bundle exec vagrant service-manager install-cli docker --cli-version 1.12.1`
```
hferentschik, line 57: Any reason we only test the _docker_ download? Where are tests for downloading _oc_?

features/support/env.rb
```diff
@@ -150,5 +150,15 @@ def extract_process_id(data)
   end
 
   expect(binaries.size).to eq(1)
-  expect(File.executable?(binaries[0])).to eq(true)
+  expect(File.executable?(binaries.first)).to eq(true)
+end
+
+Then(/^the binary "([^"]*)" should be installed with version "([^"]*)"$/) do |binary, version|
+  binaries = []
+  Find.find("#{ENV['VAGRANT_HOME']}/data/service-manager/bin") do |path|
```
hferentschik, line 158: So here you are using the data directory to test something. Why can you not write a test which uses the --path options with a target of let's say  "#{ENV['VAGRANT_HOME']}/data/service-manager/custom-path". After completion the binary should be in the custom directory, right? BTW, if the custom directory does not exist, does it get created or what happens?

hferentschik:
```
When I run `bundle exec vagrant service-manager install-cli docker --path #{ENV['VAGRANT_HOME']}/data/service-manager/foo`
And the binary "docker" should be installed with path "#{ENV['VAGRANT_HOME']}/data/service-manager/foo" (how do I verify here?)
```

Otherwise create tmpdir and let it install into the temp dir. IMO not so nice. Better to install into something which stays in the build directory.

Gemfile
brgnepal: I don't know about the environment, but it seems version `0.42.0` not working for me at least please check your version.

More details here https://github.com/projectatomic/vagrant-service-manager/pull/391#issuecomment-250703761

brgnepal: Also, CI I think install latest always unless it is pinned down to some version @hferentschik

test/vagrant-service-manager/installer_test.rb
brgnepal: @hferentschik, Ya agree. Just checking whether we are getting expected variables or not.

features/install-cli.feature
brgnepal: @hferentschik I separated that with `adb/cdk-openshift.feature` to avoid duplication.

features/support/env.rb
brgnepal: @hferentschik If custom directory doesn't exist, if fails with invalid path message

brgnepal: I don't know but `#{ENV['VAGRANT_HOME']}/data/service-manager/custom-path"` is not working in command

`bundle exec vagrant service-manager install-cli docker --path #{ENV['VAGRANT_HOME']}/data/service-manager/docker`

Gemfile
hferentschik: Hmm, I agree, rubocop version is somehow involved. I just don't understand. So looking at CU it indeed uses _0.43.0_ which is expected, since it is the latest release.

However, if I clean my gemset and execute a _bundle install_, I also should get this latest version, even w/o your change, right? But I get 0.41.2!?

test/vagrant-service-manager/installer_test.rb
hferentschik: Usually, I would refrain from testing internal state or methods (there are exceptions of cause). Much better to test the external "contract" a class with its public method forms.

Gemfile
hferentschik: ARGH, one of course has to delete the _Gemfile.lock_ as well. Otherwise even with deleting the gems, you only get the old versions back. Make sense now.

features/install-cli.feature
hferentschik: but that's just the plain _install-cli_, right? I believe in this case we don't have the _--path_ and _--version_ option, right? Something we should test to verify the expected error message occurs. On the adb feature file we still can then test the 'oc' download with path and version.

features/support/env.rb
hferentschik: >  If custom directory doesn't exist, if fails with invalid path message

Which should be put in a test. Remember, these acceptance test also describe the expected behavior of all the commands, including the error cases.

hferentschik: > I don't know but #{ENV['VAGRANT_HOME']}/data/service-manager/custom-path" is not working in command

What's the error? My guess is that the variable substitution does not work. Did you debug it? Check _aruba/cucumber/command.rb_ and _aruba/api/command.rb_. If backtick does not work, we can use _system_ and write our own step definition. Have a go at it.

hferentschik:
@budhrg afaict nothing has changed since my last review, or am I missing something?

brgnepal:
@hferentschik sorry couldn't update this PR as I am still finding the way to write test for --path option.

hferentschik:
> sorry couldn't update this PR as I am still finding the way to write test for --path option.

no worries. For some reason I thought you wanted me to have another look.

brgnepal:
@hferentschik Updated test for --path option. Used /tmp as of now. Not getting framework related path to use.

features/install-cli.feature
```diff
@@ -52,6 +54,14 @@ Feature: Command behavior of client side tools installation
...
+    Then the exit status should be 0
+    And the binary "docker" should be installed with version "1.12.1"
+
+    When I run `bundle exec vagrant service-manager install-cli docker --path /tmp/docker`
```
hferentschik: I don't think that _/tmp/docker_ is a good choice. This won't work on Windows for example. I know we cannot run tests on Windows at, but we are not so far from it. This would just put another roadblock into our way. Also what would be responsible to clean this up, aka removing already downloaded binaries?

What exactly is the issue with using a env variable here? Where does it break? Do you have a stack trace or failing test output?

hferentschik (changes requested):

.rubocop_todo.yml
```diff
@@ -13,40 +13,33 @@ Lint/AmbiguousRegexpLiteral:
...
 
 # Offense count: 1
-Lint/HandleExceptions:
+Lint/Eval:
   Exclude:
-    - 'lib/vagrant-service-manager/service.rb'
+    - 'features/support/env.rb'
 
-# Offense count: 13
+# Offense count: 17
```
hferentschik, line 20: So we are getting worse? That's not good

.rubocop_todo.yml
```diff
@@ -13,40 +13,33 @@ Lint/AmbiguousRegexpLiteral:
...
 Metrics/MethodLength:
   Max: 96
 
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 111
-  Exclude:
-    - 'lib/vagrant-service-manager/plugin_util.rb'
+  Max: 132
```
hferentschik: this will increase module length for all modules from now moving forward, opposed to exclude just _plugin_util.rb_, right?

.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2016-07-09 23:34:13 +0200 using RuboCop version 0.41.2.
+# on 2016-10-13 17:03:27 +0530 using RuboCop version 0.43.0.
```
hferentschik, line 3: For what it's worth, I did not have to make any changes to the rubocop todo file. Is this change required?

features/install-cli.feature
```diff
@@ -52,6 +54,15 @@ Feature: Command behavior of client side tools installation
     Then the exit status should be 0
     And the binary "docker" should be installed
 
+    When I run `bundle exec vagrant service-manager install-cli docker --cli-version 1.12.1`
+    Then the exit status should be 0
+    And the binary "docker" should be installed with version "1.12.1"
+
+    # '../vagrant.d' maps to ENV['VAGRANT_HOME']
+    When I run `bundle exec vagrant service-manager install-cli docker --path ../vagrant.d/docker`
```
hferentschik, line 62: So ENV['VAGRANT_HOME'] does not expand in here when used in the form of {#ENV['VAGRANT_HOME']...}?
Personally I would almost expect this to fail to begin with. Is it intentional to support relative paths?

locales/en.yml
```diff
@@ -104,7 +104,9 @@ en:
             A supported service. For example: docker, kubernetes or openshift.
 
           Options:
-                -h, --help         print this help
+                -h, --help            print this help
+                --cli-version         binary version to install
+                --path                local path where to install the binary
```
hferentschik: should probably mention whether this path is absolute or relative or both

test/vagrant-service-manager/installer_test.rb
hferentschik: @budhrg did you do something about this now? Remove?

hferentschik (changes requested):

Gemfile
hferentschik: Actually, I think this is technically not required. If you empty your gems and remove the Gemfile.lock all is good. You will automatically get 0.43. It is a bit sneaky to add this here.

.rubocop_todo.yml
brgnepal: ya correct @hferentschik . Reverting it now.

features/install-cli.feature
brgnepal: @hferentschik In this line, it is actually running as shell command, so `#{ENV['VAGRANT_HOME]}` form will not work here.

locales/en.yml
brgnepal: Changed to "absolute or relative path where to install the binary"

.rubocop_todo.yml
brgnepal: @hferentschik , I ran the command rubocop --auto-gen-config which gave me new set of offenses. I think this will help to monitor the list of offenses we are doing. 
For example, with new auto-gen-config, we found new offenses like below Abc count increased to 17 from 13.

.rubocop_todo.yml
brgnepal: @hferentschik I think this offenses has been accumulated from last we ran `rubocop --auto-gen-config`. I think we need to create an issue to do a first set of rubocop offense cleanup.

Gemfile
brgnepal: @hferentschik , Do you mean everytime, I need to remove Gemfile.lock file to map to ci like bundler environment?

features/install-cli.feature
brgnepal: @hferentschik Add testcase for oc in adb-openshift.feature file

brgnepal:
retest this please

.rubocop_todo.yml
hferentschik: Might be worse to do all this rubocop fiddling separately. It is not really related to the issue at hand at all.

features/install-cli.feature
hferentschik: well, usually backtick operator expands variables, right? Is this not the case here?

hferentschik (changes requested):

lib/vagrant-service-manager/plugin_util.rb
```diff
@@ -162,9 +162,11 @@ def self.fetch_existing_oc_binary_path_in_windows
         path_string = ENV['PATH']
 
         separator = ':' unless path_string.include?(';')
-        path_string.split(separator).detect do |dir_path|
+        oc_path_dir = path_string.split(separator).detect do |dir_path|
```
hferentschik, line 165: So what is this change about? How is it related to the issue? This is a Windows issue and not related to this added tests, right? Is there another issue on Windows which gets fixed here?

brgnepal: @hferentschik Yes, I forgot to return the detected oc.exe path. Instead I was returning the first detected `dir_path`

brgnepal:
@hferentschik Also added documentation in `commands.adoc`

commands.adoc
```diff
@@ -223,10 +223,18 @@ Similarly, you can use `$ vagrant service-manager status openshift`,
 
 If you need to install the client binary for a specified service, use: +
 
-`vagrant service-manager install-cli [service]` +
+`vagrant service-manager install-cli [service] [options] [arg]` +
 
 The possible options for service are `docker` and `openshift`. +
 
+The following options can be used: +
```
brgnepal, line 230: @Preeticp Please have a quick look to this docs.

cc @hferentschik

commands.adoc
```diff
@@ -223,10 +223,18 @@ Similarly, you can use `$ vagrant service-manager status openshift`,
...
+
+`--cli-version`::  Version of binary to install +
+`--path`::   Absolute or relative path where to install the binary +
```
Preeticp, line 233: Replace:
Version of binary to install>Displays the version of the binary to be installed.
Absolute or relative path where to install the binary>Displays the absolute or relative path where the binary should be installed.

commands.adoc
```diff
@@ -223,10 +223,18 @@ Similarly, you can use `$ vagrant service-manager status openshift`,
...
+`-h, --help`::   Prints help for the specific command being run.
+
+_arg_ represents the valid value for the respective `--cli-version` and/or `--path` options.
+
```
Preeticp, line 237: This is the only place we are using Italics for words like '_arg_'. I would prefer we use `[arg]`/`arg` to keep it consistent with the style and also to prevent confusion for users. However, whichever style we decide to use, it would have to be applied to the entire document for words like, commands, sub-commands, options etc. Or we can simply change _arg_ to arg.

Preeticp, line 237: Also, should we modify the example below to show the --cli version and --path options you have added?

commands.adoc
hferentschik: @Preeticp "Displays"!? This commands let you specify a version and a path, they are not displaying them

hferentschik:
retest

hferentschik:
Seems that we have Rubocop failures again :-( This time a Rubocop 0.44 got releases. I guess we need to thing about, whether we want to pin the Rubocop version to prevent this to happen or whether we just handle it as it comes. Either way, let's do this separately. Let's create an issue and pull request for upgrading (and potentially pinning version) of Rubocop. Once this is in place we can place this pull request on top. @budhrg Any strong feeling regarding setting a fixed version number vs dealing with new "violations"?

hferentschik:
@budhrg I am also closing this pull request. Let's split Rubocop from install-cli and create new pull requests. We take it from there.

commands.adoc
brgnepal: Ya @Preeticp , Probably should use "Specify version of binary to install" and "Specify absolute or relative path where to install .."

Preeticp: Yes, Specify the version of the binary to be installed. and Specify the absolute or relative path where the binary should be installed.

brgnepal:
> feeling regarding setting a fixed version number 

@hferentschik 

I am in favor of pinning to fix version and update it later on instead of dealing with new violations every time new version is released.

brgnepal:
Also, will give the PR of this issue https://github.com/projectatomic/vagrant-service-manager/issues/412 first @hferentschik

PR #395, opened 2016-09-28
Fix #394 Forcing used Bundler version to 1.12.5 (as required by Vagra…

### LINKED ISSUE #394: Unable to bootstrap build from an empty gemset
When bootstrapping the project from an empty gemset via `bundle install` the following error occurs (on CI as well as locally):

```
[!] There was an error parsing `Gemfile`: You cannot specify the same gem twice     coming from different sources.
You specified that vagrant-service-manager (>= 0) should come from source at `.` and     source at `.`
. Bundler cannot continue.

 #  from /Users/hardy/work/developer-platform/vagrant/vagrant-service-manager/    Gemfile:27
 #  -------------------------------------------
 #  group :plugins do
 >    gem 'vagrant-service-manager', path: '.'
 #  end
 #  -------------------------------------------
```

This is with bundler 1.13.1. Might be due to a newer version of bundler or due to some changes we did to the gemspec/Gemfile configuration. Needs investigation.

hferentschik:
…nt. Also including gemspec in the plugin group of the Gemfile to avoid duplicate gem warning.

### DIFF
--- a/.ci/jenkins-execute-script.py
+++ b/.ci/jenkins-execute-script.py
@@ -64,7 +64,7 @@ for host in data['hosts']:
 
     # setup the environment
     setup_cmd  = 'cd vagrant-service-manager && '
-    setup_cmd += 'gem install bundler && '
+    setup_cmd += 'gem install bundler -v 1.12.5 && '
     setup_cmd += 'bundle install --no-color'
     execute_on_host(host, setup_cmd, "Unable to setup Ruby environment")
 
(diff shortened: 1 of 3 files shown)


hferentschik:
retest this please

hferentschik:
This fixes the Bundler issues and the environment at least bootstraps. But I see various test failures as well, partly in the feature tests as well as in the unit tests. Let's try to address this in subsequent pull requests.

hferentschik:
Merged

PR #398, opened 2016-09-28
Fix #397 Using a guard clause instead of wrapping the code inside a c…

### LINKED ISSUE #397: Rubocop failures in binary_handler.rb and installer.rb
```
Running RuboCop...
Inspecting 47 files

Offenses:
CI fails due to Rubocop failures in _binary_handler.rb_ and _installer.rb_

[36mlib/vagrant-service-manager/binary_handlers/binary_handler.rb[0m:103:9: [33mC[0m: Use a guard clause instead of wrapping the code inside a conditional expression.
    unless @options['--cli-version'] != oc_version || @options['--path']
    ^^^^^^
[36mlib/vagrant-service-manager/installer.rb[0m:32:9: [33mC[0m: Use a guard clause instead of wrapping the code inside a conditional expression.
    unless PluginUtil.service_running?(@machine, @type.to_s)
    ^^^^^^

47 files inspected, [31m2 offenses[0m detected
RuboCop failed!
```

The violations look legit, but I don't seem to get them on my local dev machine!?

hferentschik:
…onditional

### DIFF
--- a/lib/vagrant-service-manager/binary_handlers/binary_handler.rb
+++ b/lib/vagrant-service-manager/binary_handlers/binary_handler.rb
@@ -100,10 +100,9 @@ module VagrantPlugins
           oc_version = CDKOpenshiftBinaryHandler::LATEST_OC_VERSION
         end
 
-        unless @options['--cli-version'] != oc_version || @options['--path']
-          path = PluginUtil.fetch_existing_oc_binary_path_in_windows
-          @path = path unless path.nil?
-        end
+        return if @options['--cli-version'] != oc_version || @options['--path']
+        path = PluginUtil.fetch_existing_oc_binary_path_in_windows
+        @path = path unless path.nil?
       end
 
       def archive_dir_name
--- a/lib/vagrant-service-manager/installer.rb
+++ b/lib/vagrant-service-manager/installer.rb
@@ -29,11 +29,9 @@ module VagrantPlugins
       private
 
       def validate_prerequisites
-        unless PluginUtil.service_running?(@machine, @type.to_s)
-          @env.ui.info I18n.t('servicemanager.commands.install_cli.service_not_enabled',
-                              service: @type)
-          exit 126
-        end
+        return if PluginUtil.service_running?(@machine, @type.to_s)
+        @env.ui.info I18n.t('servicemanager.commands.install_cli.service_not_enabled', service: @type)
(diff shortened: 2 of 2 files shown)


lib/vagrant-service-manager/binary_handlers/binary_handler.rb
```diff
@@ -100,10 +100,9 @@ def handle_windows_binary_path
           oc_version = CDKOpenshiftBinaryHandler::LATEST_OC_VERSION
         end
 
-        unless @options['--cli-version'] != oc_version || @options['--path']
-          path = PluginUtil.fetch_existing_oc_binary_path_in_windows
-          @path = path unless path.nil?
-        end
+        return if @options['--cli-version'] != oc_version || @options['--path']
```
hferentschik, line 103: The change just addresses the Rubocop violation. I am a bit confused what's going on in this code otherwise? What's the purpose of the method and how is this working. Is the indent to not update @path if certain conditions are not met? Where are the error messages then? To me this code looks fishy.

hferentschik: @budhrg !?

hferentschik:
@budhrg Seems my two build requests in combination get the build working again. Can I go ahead and merge?

lib/vagrant-service-manager/binary_handlers/binary_handler.rb
brgnepal: @hferentschik , Code looks fine. 
The idea is look for windows existing paths and set if found like `C:\DevSuite\bin\oc.exe`. 

The path should not be set if the user provided `--cli-version` is different than pinned oc version or user mentioned `--path` option.

brgnepal:
LGTM :+1:

brgnepal:
Merged and thanks.

lib/vagrant-service-manager/binary_handlers/binary_handler.rb
hferentschik: Right, you know that and it might work like this, but the code works heavily with side effects. If it completes the path is set, otherwise I take it that is exits earlier and somewhere else something is a check against _nil_ or so. I am not saying that it is not working. What I am saying is that it is not written in a clear way. Someone looking at it for the first time wonders about the intentions here.

Issue #396, opened 2016-09-28
Test failures due to SSL verification error


hferentschik:
```
# Running:

......................FEF............

Finished in 8.245352s, 4.4874 runs/s, 9.5812 assertions/s.

  1) Failure:
VagrantPlugins::ServiceManager::ADBKubernetesBinaryHandler#test_0004_should raise     error with invalid --cli-version [/Users/hardy/work/developer-platform/vagrant/    vagrant-service-manager/test/vagrant-service-manager/binary_handlers/    adb_kubernetes_binary_handler_test.rb:54]:
[VagrantPlugins::ServiceManager::URLValidationError] exception expected, not
Class: <OpenSSL::SSL::SSLError>
Message: <"SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B:     certificate verify failed">
---Backtrace---
/Users/hardy/.rvm/rubies/ruby-2.0.0-p643/lib/ruby/2.0.0/net/http.rb:921:in `connect'
/Users/hardy/.rvm/rubies/ruby-2.0.0-p643/lib/ruby/2.0.0/net/http.rb:921:in `block in     connect'
/Users/hardy/.rvm/rubies/ruby-2.0.0-p643/lib/ruby/2.0.0/timeout.rb:52:in `timeout'
/Users/hardy/.rvm/rubies/ruby-2.0.0-p643/lib/ruby/2.0.0/net/http.rb:921:in `connect'
/Users/hardy/.rvm/rubies/ruby-2.0.0-p643/lib/ruby/2.0.0/net/http.rb:862:in `do_start'
/Users/hardy/.rvm/rubies/ruby-2.0.0-p643/lib/ruby/2.0.0/net/http.rb:851:in `start'
/Users/hardy/.rvm/rubies/ruby-2.0.0-p643/lib/ruby/2.0.0/net/http.rb:1373:in `request'
/Users/hardy/.rvm/rubies/ruby-2.0.0-p643/lib/ruby/2.0.0/net/http.rb:1297:in     `request_head'
/Users/hardy/work/developer-platform/vagrant/vagrant-service-manager/lib/    vagrant-service-manager/binary_handlers/binary_handler.rb:139:in `validate_url'
/Users/hardy/work/developer-platform/vagrant/vagrant-service-manager/test/    vagrant-service-manager/binary_handlers/adb_kubernetes_binary_handler_test.rb:54:in     `block (3 levels) in <module:ServiceManager>'
---------------


[...]

PR #399, opened 2016-09-30
Fix #393 Kubernetes acceptance tests fail on OS X

### LINKED ISSUE #393: Kubernetes acceptance tests fail on OS X
Follow up on this commit 565d25c66c736a4bfd5eaf988c76424998bad252

```
 3) Failure:
VagrantPlugins::ServiceManager::ADBKubernetesBinaryHandler#test_0002_should build download url [/Users/hardy/work/developer-platform/vagrant/vagrant-service-manager/test/vagrant-service-manager/binary_handlers/adb_kubernetes_binary_handler_test.rb:42]:
--- expected
+++ actual
@@ -1 +1 @@
-"https://storage.googleapis.com/kubernetes-release/release/v1.2.0/bin/linux/amd64/kubectl"
+"https://storage.googleapis.com/kubernetes-release/release/v1.2.0/bin/darwin/amd64/kubectl"
```

Seems to be again a OS specific expectation here.

brgnepal:
Fix #393 

Looks like I miss to update the regex term

### DIFF
--- a/test/vagrant-service-manager/binary_handlers/adb_kubernetes_binary_handler_test.rb
+++ b/test/vagrant-service-manager/binary_handlers/adb_kubernetes_binary_handler_test.rb
@@ -36,7 +36,7 @@ module VagrantPlugins
 
       it 'should build download url' do
         expected_url = @base_download_url + '/v1.2.0/bin/linux/amd64/kubectl'
-        expected_url.sub!(/Linux/, 'Darwin') if Vagrant::Util::Platform.darwin?
+        expected_url.sub!(/linux/, 'darwin') if Vagrant::Util::Platform.darwin?
 
         @handler.send(:build_download_url)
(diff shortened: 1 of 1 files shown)


brgnepal:
@hferentschik The CI is green. :+1:

hferentschik:
Merged, thanks. Now I have only issue #396 left to move on ;-) @budhrg, I take it that you don't see this SSL issue, right? Seems CI does not see it either. There are apparently various reasons for this error, among others invalid certificates on my machine. I just don't understand why it is happening.

@budhrg, BTW, it is a bit unfortunate that the tests need an internet connection to work. IMO we should stub the HTTP connection part. No need to make an actual live connection.

brgnepal:
Ya @hferentschik , I couldn't able to reproduce that SSL error.

> we should stub the HTTP connection part. No need to make an actual live connection. 

What about box checks?

Issue #362, opened 2016-08-01
Document how to configure OpenShift with custom image tag, registry etc 


brgnepal:
Hi,

It seems the documentation on configuring OpenShift with custom image tag, registry has be removed by mistake. Unable to find the doc from past five minutes in both here and ADB repo.

May be its there but its quite not visible properly.

It looks like something this mentioned in this [quick start of release v1.0.0](https://github.com/projectatomic/vagrant-service-manager/tree/v1.0.0#quick-start) 3rd and 4th points.

brgnepal:
cc @Preeticp

Preeticp:
@budhrg this documentation exists in ADB usage document, in the using ADB with host based tools section, as discussed earlier.

brgnepal:
Found the docs [here](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/using.rst#using-the-adb-with-host-based-tools-eclipse-and-clis) as we decided to move it there. This will be more important in customizing OpenShift I would suggest here to make a reference of it in VSM repo. Something inside the sections like configurations etc etc and then point to this docs.

brgnepal:
@bexelbie I think this feature is getting hidden in some sense since it is now not visible properly in VSM and nor in ADB README directly. 

Even inside ADB, it is only mentioned in [using.rst](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/using.rst) page under section "Using ADB with host-based tools (Eclipse and CLIs) # 3".

Is there any way to make such feature visible?
Like we can give users all options they can try in separate docs.

Preeticp:
+1 @budhrg for giving users all the options they can try. Could you please mention which options you would like to cover and the relevant details, based on that, a decision can be arrived at, as to whether we need different sections for those options within Usage doc or we need separate docs for each of the option. Also depending on what needs to be covered we may want to create a separate issue for this in ADB repo, and linking VSM repo to those docs/sections can be done at the same time.

hferentschik:
@budhrg, just to be clear - you are talking about these options, right?
- config.servicemanager.openshift_docker_registry 
- config.servicemanager.openshift_image_name 
- config.servicemanager.openshift_image_tag 

[...]

PR #400, opened 2016-09-30
Fix #302 service-manager config options for HTTP proxy

### LINKED ISSUE #302: Provide service-manager config options for HTTP proxy settings
Since we are working with proxy issue and it would be nice that service-manager have variable to consume it like it have for openshift image, tag and docker registry.

brgnepal:
Fix #302

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -152,6 +152,23 @@ code uses the last IPv4 address available from the set of configured
 addresses that are _up_. i.e. if eth0, eth1, and eth2 are all up and
 have IPv4 addresses, the address on eth2 is used.
 
+=== HTTP Proxy Settings
+
+In an environment where HTTP traffic needs pass through a HTTP proxy, Docker
+and OpenShift need to be aware of the required proxy settings in order to function.
+`http_proxy`, `http_proxy_user` and `http_proxy_password` configurations need to be
+specified through Vagrantfile. You can do so via:
+
+----------------------------------------------------------------------------
+config.servicemanager.http_proxy = <Proxy URL>
+config.servicemanager.http_proxy_user = <Proxy user name>
+config.servicemanager.http_proxy_password = <Proxy user password>
+----------------------------------------------------------------------------
+
+When these settings are applied, they are passed through to the Docker and
+OpenShift service. `http_proxy_user` and `http_proxy_password` can be omitted in an
+unauthenticated proxy environment.
+
 == Development
 
 === Setup
--- a/lib/vagrant-service-manager/service_base.rb
+++ b/lib/vagrant-service-manager/service_base.rb
@@ -5,9 +5,9 @@ module VagrantPlugins
         @machine = machine
         @env = env
         @ui = env.respond_to?('ui') ? env.ui : env[:ui]
+        @services = @machine.config.servicemanager.services.split(',').map(&:strip)
         home_path = env.respond_to?('home_path') ? env.home_path : env[:home_path]
         @plugin_dir = File.join(home_path, 'data', 'service-manager')
-        @services = @machine.config.servicemanager.services.split(',').map(&:strip)
       end
 
       def service_start_allowed?
@@ -16,6 +16,43 @@ module VagrantPlugins
 
       def cdk?
         @machine.guest.capability(:os_variant) == 'cdk'
+      end
+
+      def proxy_cmd_options
+        options = ''
+
+        return options unless http_proxy_settings_valid?
+
+        PROXY_CONFIG.each do |key|
+          options += "#{key.to_s.upcase}='#{@machine.config.servicemanager.send(key)}' "
+        end
+
+        options.chop
+      end
+
+      def http_proxy_settings_valid?
+        proxy = @machine.config.servicemanager.send('http_proxy')
+        user = @machine.config.servicemanager.send('http_proxy_user')
+        password = @machine.config.servicemanager.send('http_proxy_password')
+
+        if proxy && user.nil? && password.nil? || (proxy && user && password)
+          PluginLogger.debug('Detected proxy settings. Going to apply them to service commands.')
+          values = "Proxy URL : #{proxy}"
+          values += ", User : #{user}, Password: *** " unless user.nil?
+          PluginLogger.debug(values)
+          return true
+        end
+
+        warn_proxy_settings_missing(proxy, user, password)
+        false
+      end
+
+      def warn_proxy_settings_missing(proxy, user, password)
+        # Check if user try to set proxy
+        return if proxy.nil? && user.nil? && password.nil?
+        message = 'Proxy URL is missing' if proxy.nil?
+        message = 'Either user or password is missing.' if proxy && user.nil? || proxy && password.nil?
+        @ui.warn message
       end
     end
   end
--- a/lib/vagrant-service-manager/services/open_shift.rb
+++ b/lib/vagrant-service-manager/services/open_shift.rb
@@ -6,12 +6,14 @@ module VagrantPlugins
       def initialize(machine, env)
         super(machine, env)
         @service_name = 'openshift'
-        @extra_cmd = build_extra_command
       end
 
       def execute
         if service_start_allowed?
-          command = "#{@extra_cmd} sccli openshift"
+          command = 'sccli openshift'
+          command = "#{extra_command_options} #{command}" unless extra_command_options.empty?
+          command = "#{proxy_cmd_options} #{command}" unless proxy_cmd_options.empty?
+
           PluginUtil.execute_and_exit_on_fail(@machine, @ui, command)
         end
       rescue Vagrant::Errors::GuestCapabilityNotFound
@@ -47,14 +49,16 @@ module VagrantPlugins
 
       private
 
-      def build_extra_command
+      def extra_command_options
         cmd = ''
-        CONFIG_KEYS.select { |e| e[/^openshift_/] }.each do |key|
+
+        OPENSHIFT_CONFIG.each do |key|
           unless @machine.config.servicemanager.send(key).nil?
             env_name = key.to_s.gsub(/openshift_/, '').upcase
             cmd += "#{env_name}='#{@machine.config.servicemanager.send(key)}' "
           end
         end
+
         cmd.chop
       end
 
--- a/test/vagrant-service-manager/service_base_test.rb
+++ b/test/vagrant-service-manager/service_base_test.rb
@@ -1,0 +1,66 @@
+require_relative '../test_helper'
+require 'vagrant-service-manager/config'
+require 'vagrant-service-manager/plugin_util'
+
+module VagrantPlugins
+  describe ServiceManager::ServiceBase do
+    let(:machine) { fake_machine }
+    let(:instance) { ServiceManager::ServiceBase.new(machine, machine.env) }
+
+    it 'should build the proxy options if only proxy url is given' do
+      machine.config.servicemanager.http_proxy = 'foo:8080'
+      options = instance.proxy_cmd_options
+      assert_match(/HTTP_PROXY='foo:8080'/, options)
+    end
+
+    it 'should build the proxy options if all proxy settings are given' do
+      machine.config.servicemanager.http_proxy = 'foo:8080'
+      machine.config.servicemanager.http_proxy_user = 'user'
+      machine.config.servicemanager.http_proxy_password = 'password'
+
+      options = instance.proxy_cmd_options
+      assert_match(/HTTP_PROXY='foo:8080'/, options)
+      assert_match(/HTTP_PROXY_USER='user'/, options)
+      assert_match(/HTTP_PROXY_PASSWORD='password'/, options)
+    end
+
+    it 'should return empty proxy options if proxy url is not given' do
+      machine.config.servicemanager.http_proxy_user = 'user'
+      machine.config.servicemanager.http_proxy_password = 'password'
+
+      options = instance.proxy_cmd_options
+      assert_match(options, '')
+    end
+
+    it 'should return empty proxy options if user name not given' do
+      machine.config.servicemanager.http_proxy = 'foo:8080'
+      machine.config.servicemanager.http_proxy_password = 'password'
+
+      options = instance.proxy_cmd_options
+      assert_match(options, '')
+    end
+
+    it 'should return empty proxy options if password not given' do
+      machine.config.servicemanager.http_proxy = 'foo:8080'
+      machine.config.servicemanager.http_proxy_user = 'user'
+
+      options = instance.proxy_cmd_options
+      assert_match(options, '')
+    end
+
+    it 'should pass proxy settings as specified via Vagrant config' do
+      machine.config.servicemanager.http_proxy = 'foo:8080'
+      machine.config.servicemanager.http_proxy_user = 'user'
+      machine.config.servicemanager.http_proxy_password = 'password'
+      command = "#{instance.proxy_cmd_options} sccli openshift"
+
+      ServiceManager::PluginUtil.execute_once(machine, FakeUI, command)
+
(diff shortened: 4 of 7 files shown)


hferentschik:
What happened to testing?

brgnepal:
@hferentschik Not sure how to add acceptance tests for this (Proxy checks)

hferentschik:
Add unit tests

test/vagrant-service-manager/service_base_test.rb
```diff
@@ -0,0 +1,22 @@
+require_relative '../test_helper'
+require 'vagrant-service-manager/config'
+require 'vagrant-service-manager/plugin_util'
+
+module VagrantPlugins
+  describe ServiceManager::ServiceBase do
+    let(:machine) { fake_machine }
+    let(:instance) { ServiceManager::ServiceBase.new(machine, machine.env) }
+
+    it 'should build the proxy options for command properly' do
```
brgnepal: @hferentschik I couldn't write the similar tests as [this - `rhn_register_test.rb`](https://github.com/projectatomic/adb-vagrant-registration/blob/master/test/vagrant-registration/cap/redhat/rhn_register_test.rb#L37-L46) you mentioned.
The reason I found that our code to execute command i.e [`PluginUtil.execute_and_exit_on_fail`](https://github.com/projectatomic/vagrant-service-manager/blob/master/lib/vagrant-service-manager/plugin_util.rb#L93) contains more stuff then just executing command like in case of `rhn_register_test.rb`. 

In our case, the method fails on exit_code being empty string and undefined method `zero?` which I found same in `rhn_register_test.rb` where the exit_code is emptry string.

brgnepal:
@hferentschik Added a unit test to check the functionality of method which is responsible for building proxy options for commands.

test/vagrant-service-manager/service_base_test.rb
hferentschik: > I couldn't write the similar tests as this - rhn_register_test.rb you mentioned.
> The reason I found that our code to execute command i.e PluginUtil.execute_and_exit_on_fail contains more stuff then just executing command like in case of rhn_register_test.rb. 

Why not, do you have an example of what you tried to do?

> In our case, the method fails on exit_code being empty string and undefined method zero? which I found same in rhn_register_test.rb where the exit_code is emptry string.

But you can return whatever you want, right? the communicator should be stubbed. I think it would help to see what you tried to do.

lib/vagrant-service-manager/service_base.rb
```diff
@@ -17,6 +17,18 @@ def service_start_allowed?
       def cdk?
         @machine.guest.capability(:os_variant) == 'cdk'
       end
+
+      def proxy_cmd_options
+        options = ''
+
+        PROXY_CONFIG.each do |key|
+          unless @machine.config.servicemanager.send(key).nil?
```
hferentschik: Is it worth to verify that all options are specified or none?

hferentschik: Maybe it also makes sense to add some info/debug trace here - "Detected proxy settings. Going to apply them to service commands" Something along these lines. If not as info, at least as debug. Also it might make sense to add the options verification and logging to initialize, instead of doing it in here.

lib/vagrant-service-manager/services/open_shift.rb
```diff
@@ -6,12 +6,14 @@ class OpenShift < ServiceBase
       def initialize(machine, env)
         super(machine, env)
         @service_name = 'openshift'
-        @extra_cmd = build_extra_command
       end
 
       def execute
         if service_start_allowed?
-          command = "#{@extra_cmd} sccli openshift"
+          command = 'sccli openshift'
```
hferentschik, line 13: Out of interest, why do you use string interpolation here. Would it not be much easier to use plain string concatenation. IMO it would also be a bit clearer.

On a different note, what do you think about renaming _open_shift.rb_ to _openshift.rb_. This underscore keeps nagging me.

brgnepal: @hferentschik This underscore help to dynamically build the OpenShift class name. Otherwise it would difficult to do so with 'openshift' where there is no separation.

lib/vagrant-service-manager/service_base.rb
brgnepal: Good idea :+1:

LalatenduMohanty:
@budhrg can you please explain how will be the user experience i.e. how user is going to use proxy information with VSM? Also I do not see any documentation change for this PR?

brgnepal:
@LalatenduMohanty The user will able to leverage this via configurations as:

```
config.servicemanager.http_proxy = '10.70.49.109:3128'
config.servicemanager.http_proxy_user = 'fedora'
config.servicemanager.http_proxy_password = 'xxxx'
config.servicemanager.services = 'docker, openshift'
```

I think documentation, we can add later on since for these configurations only I created one issue which still need to be addressed. https://github.com/projectatomic/vagrant-service-manager/issues/362

LalatenduMohanty:
@budhrg  cool.  I think #362 is a different issue. You should create another issue for this.

lib/vagrant-service-manager/services/open_shift.rb
hferentschik: > This underscore help to dynamically build the OpenShift class name. Otherwise it would difficult to do so with 'openshift' where there is no separation.

What do you mean? Are you saying that the file name as a bearing on how some code behaves? Then I give an additional vote to renaming to _openshift.rb_. Relying on the filename is wrong.

hferentschik:
> @budhrg can you please explain how will be the user experience i.e. how user is going to use proxy information with VSM? Also I do not see any documentation change for this PR?

Very good point. The doc side of things is completely missing.

hferentschik:
> I think documentation, we can add later on since for these configurations 

-1 - A new feature/fix consists of: the fix, tests and required doc changes. Why would you tackle doc changes somewhere else than here?

brgnepal:
@hferentschik @LalatenduMohanty Issue https://github.com/projectatomic/vagrant-service-manager/issues/401 has been created to track it.

lib/vagrant-service-manager/services/open_shift.rb
brgnepal: > Then I give an additional vote to renaming to openshift.rb. Relying on the filename is wrong.

This is general pattern in Ruby to have such names. Ruby on Rails codebase is full of this pattern which follows convention over configurations. It would have been difficult if they didn't use this pattern.

test/vagrant-service-manager/service_base_test.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    it 'should fail building proxy options if password not given' do
+      machine.config.servicemanager.http_proxy = 'foo:8080'
+      machine.config.servicemanager.http_proxy_user = 'user'
+
+      options = instance.proxy_cmd_options
+      assert_match(options, '')
+    end
+
+    it 'should pass proxy settings as specified via Vagrant config' do
```
brgnepal, line 51: @hferentschik Similar to vagrant-registration test.

brgnepal:
@hferentschik Added few more testcases related to proxy validations and test similar to vagrant-registration to capture command being fire.

lib/vagrant-service-manager/services/open_shift.rb
hferentschik: This is not Rails, right? Having a coupling of file name to internal code is dangerous. So where exactly are you making use of this?

hferentschik:
> Added few more testcases related to proxy validations and test similar to vagrant-registration to capture command being fire.

I'll take a look

lib/vagrant-service-manager/services/open_shift.rb
brgnepal: @hferentschik Created issue https://github.com/projectatomic/vagrant-service-manager/issues/402 to track it.

README.adoc
```diff
@@ -127,6 +127,19 @@ NOTE: Debug output from `vagrant-service-manager` is prepended with
 the following string:
 `DEBUG command: [ service-manager: <command name / log message> ]`
 
+=== Configurations
```
brgnepal: @hferentschik Added docs here.

hferentschik: I think I'd just call "HTTP Proxy configuration" and place it after "Ip address detection"

README.adoc
```diff
@@ -127,6 +127,19 @@ NOTE: Debug output from `vagrant-service-manager` is prepended with
...
+
+- HTTP Proxy Settings +
+User can able to start ADB in proxy environment by specifying following configurations
```
hferentschik: I think you need to be a bit more verbose and explain the context. Something like:
"In a environment where HTTP traffic needs pass through a HTTP proxy, Docker and OpenShift need to be aware of the required proxy settings in order to function. _http_proxy_, _http_proxy_user_ and _http_proxy_password_ need to be specified. You can do so via:

README.adoc
```diff
@@ -127,6 +127,19 @@ NOTE: Debug output from `vagrant-service-manager` is prepended with
...
+through Vagrantfile as:
+
++
+----------------------------------------------------------------------------
+config.servicemanager.http_proxy = <Proxy URL>
+config.servicemanager.http_proxy_user = <Proxy user name>
+config.servicemanager.http_proxy_password = <Proxy user password>
+----------------------------------------------------------------------------
+
```
hferentschik: When these settings are applied, they are passed through to the Docker and OpenShift service. _http_proxy_user_ and _http_proxy_password can be omitted in an unauthenticated proxy environment

hferentschik: Does this work on restart? What I mean is, can I start the VM w/o these settings, then add them and then call _vagrant reload_ and it will work?

brgnepal: I think on vagrant reload, it will not work since we are not invoking our hooks on it. Probably we need to add it. When add it, it means it would be a separate feature since I don't remember we have taken care of this (vagrant reload) use case.

README.adoc
brgnepal: > I think I'd just call "HTTP Proxy configuration" and place it after "Ip address detection"

I kept it so that we can add more Vagrantfile configurations specifics docs in this section only. 
Like the openshift options configurations which are now only documented under [Using ADB with host-based tools](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/using.adoc#using-adb-with-host-based-tools-eclipse-and-clis) point 3.

lib/vagrant-service-manager/service_base.rb
```diff
@@ -17,6 +17,43 @@ def service_start_allowed?
...
+
+      def validation_success?
+        proxy = @machine.config.servicemanager.send('http_proxy')
+        user = @machine.config.servicemanager.send('http_proxy_user')
+        password = @machine.config.servicemanager.send('http_proxy_password')
+
+        if proxy && user.nil? && password.nil? || (proxy && user && password)
+          PluginLogger.debug('Detected proxy settings. Going to apply them to service commands.')
+          values = "Proxy URL : #{proxy}"
+          values += ", User : #{user}, Password: #{password}" unless user.nil?
```
hferentschik: I like it. However, I think we should not echo the password plain text. The settings might come via environment variables to avoid displaying the password anywhere. Yet this would allow to dump it onto the console. Maybe just "***"!?

lib/vagrant-service-manager/service_base.rb
```diff
@@ -17,6 +17,43 @@ def service_start_allowed?
...
+        return options unless validation_success?
+
+        PROXY_CONFIG.each do |key|
+          options += "#{key.to_s.upcase}='#{@machine.config.servicemanager.send(key)}' "
+        end
+
+        options.chop
+      end
+
+      def validation_success?
```
hferentschik: I'd call this _http_proxy_settings_valid?_ or something which is more specific

lib/vagrant-service-manager/service_base.rb
brgnepal: @hferentschik ok. seems fine

test/vagrant-service-manager/service_base_test.rb
```diff
@@ -0,0 +1,66 @@
+require_relative '../test_helper'
+require 'vagrant-service-manager/config'
+require 'vagrant-service-manager/plugin_util'
+
+module VagrantPlugins
```
hferentschik, line 5: +1 Nice. I like :-)

test/vagrant-service-manager/service_base_test.rb
```diff
@@ -0,0 +1,66 @@
...
+      machine.config.servicemanager.http_proxy_user = 'user'
+      machine.config.servicemanager.http_proxy_password = 'password'
+
+      options = instance.proxy_cmd_options
+      assert_match(/HTTP_PROXY='foo:8080'/, options)
+      assert_match(/HTTP_PROXY_USER='user'/, options)
+      assert_match(/HTTP_PROXY_PASSWORD='password'/, options)
+    end
+
+    it 'should fail building proxy options if proxy url is not given' do
```
hferentschik: The test is ok, but I think the expectation is wrongly formulated. The building of the proxy option is not failing. Failing would imply a exception or something. Is the expectation not more "should return empty proxy options if proxy url is not given"

test/vagrant-service-manager/service_base_test.rb
```diff
@@ -0,0 +1,66 @@
...
+      machine.config.servicemanager.http_proxy_user = 'user'
+      machine.config.servicemanager.http_proxy_password = 'password'
+
+      options = instance.proxy_cmd_options
+      assert_match(options, '')
+    end
+
+    it 'should fail building proxy options if user name not given' do
```
hferentschik: same as above.

test/vagrant-service-manager/service_base_test.rb
```diff
@@ -0,0 +1,66 @@
...
+      machine.config.servicemanager.http_proxy = 'foo:8080'
+      machine.config.servicemanager.http_proxy_password = 'password'
+
+      options = instance.proxy_cmd_options
+      assert_match(options, '')
+    end
+
+    it 'should fail building proxy options if password not given' do
```
hferentschik: same as above

test/vagrant-service-manager/service_base_test.rb
hferentschik: +1

README.adoc
hferentschik: _Configurations_ still does not mean anything to me. For now, I'd just skip this level. Let's pull 'HTTP Proxy Settings' up to the level of 'Configurations' and remove it.

README.adoc
```diff
@@ -127,6 +127,26 @@ NOTE: Debug output from `vagrant-service-manager` is prepended with
...
+config.servicemanager.http_proxy = <Proxy URL>
+config.servicemanager.http_proxy_user = <Proxy user name>
+config.servicemanager.http_proxy_password = <Proxy user password>
+----------------------------------------------------------------------------
++
+
+When these settings are applied, they are passed through to the Docker and
+OpenShift service. `http_proxy_user` and `http_proxy_password` can be omitted in an
+unauthenticated proxy environment.
+
```
hferentschik: What is the answer to my question regarding _vagrant reload_?

hferentschik (commented):
Looking good. Only some minor comments left.

test/vagrant-service-manager/service_base_test.rb
brgnepal: Ya. I mean that only. Will update it.

README.adoc
brgnepal: @hferentschik Will create a new issue since we have not been considering that use case properly. Even current hook behavior works or not I don't know ATM. I think it will not since we attached to `vagrant up` action.

hferentschik:
Merged. Thanks

PR #405, opened 2016-10-07
Fix #402 rename open_shift.rb file to openshift

### LINKED ISSUE #402: Rename open_shift.rb file to openshift.rb
Rename [`open_shift.rb`](https://github.com/projectatomic/vagrant-service-manager/blob/master/lib/vagrant-service-manager/services/open_shift.rb) to `openshift.rb`.

Make sure it doesn't add any regression.

brgnepal:
Fix #402

Looks like file name `open_shift` is not being used now as such thing is handled through hash [`SERVICE_MAP`](https://github.com/projectatomic/vagrant-service-manager/blob/master/lib/vagrant-service-manager/command.rb#L27).


hferentschik:
Rebased and merged. Thanks @budhrg for humoring me on this one. It has been nagging me for a long time. Looks so much better now :-)

PR #411, opened 2016-10-13
Fix #409 fixes Openshift service failed message on vagrant up

### LINKED ISSUE #409: Openshift service failed message on vagrant up although it is running after vagrant up process
The openshift service can be started through the plugin using the configuration as:

```
config.servicemanager.services = "openshift"
```

However, using so, I am getting wrong message as "Openshift service failed to configure properly..." even though the status of openshift service is "running" after "vagrant up" process.

**Steps to Reproduce**
- Use the Vagrantfile from [ADB](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/components/rhel/rhel-ose/Vagrantfile)
- Update the [line 72](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/components/rhel/rhel-ose/Vagrantfile#L72) as `config.servicemanager.services = 'openshift'`
- Remove these [three lines](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/components/rhel/rhel-ose/Vagrantfile#L74-L77)

brgnepal:
Fix #409


hferentschik (changes requested):
So what am I looking at here? What am I looking for? This is about some startup message right? Can we not modify our existing _adb_openshift.feature_ to check the output of 'vagrant up', verifying the unwanted output is not there and replaced by whatever is expected. We don't even have to write a new scenario, use add another step to an existing one, right?

brgnepal:
@hferentschik 

> So what am I looking at here? What am I looking for? This is about some startup message right? Can we not modify our existing adb_openshift.feature to check the output of 'vagrant up', verifying the unwanted output is not there and replaced by whatever is expected. We don't even have to write a new scenario, use add another step to an existing one, right?

Added testcase. Thanks

features/adb-openshift.feature
```diff
@@ -24,7 +24,13 @@ Feature: Command output from various OpenShift related commands in ADB
     """
 
     When I successfully run `bundle exec vagrant up --provider <provider>`
-    And I successfully run `bundle exec vagrant service-manager env openshift`
+    Then stdout from "bundle exec vagrant up --provider <provider>" should contain:
```
hferentschik, line 27: 👍

hferentschik (approved):
Looking good. Let me try this locally. If it works I rebase and merge.

hferentschik:
@budhrg actually, this pull request does not rebase cleanly. Could you rebase yourself and push an update?

brgnepal:
@hferentschik Rebased and pushed :smile:

hferentschik:
Merged, thanks.

PR #413, opened 2016-10-14
Fix #412 Pin the rubocop version and minimize the offenses

### LINKED ISSUE #412: Pin the rubocop version and minimize the offenses
We are facing new style guide violations as so as new version of rubocop gem is released. 

Pin the rubocop to latest stable version and minimize the offenses as much as possible.

brgnepal:
- Fine tune rubocop exclusion files


.rubocop_todo.yml
```diff
@@ -1,57 +1,7 @@
...
 # `rubocop --auto-gen-config`
-# on 2016-07-09 23:34:13 +0200 using RuboCop version 0.41.2.
+# on 2016-10-14 23:34:08 +0530 using RuboCop version 0.44.1.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
-
-# Offense count: 3
-Lint/AmbiguousRegexpLiteral:
```
hferentschik, line 10: Any reason why you moved everything into  .rubocop.yml ? I guess the idea is to define in _.rubocop.yml_ basic changes to the default values, whereas in _.rubocop_todo.yml_ you keep exclusions which you are temporarily excluding until you can fix it. Would you agree? If so, are you saying the exclusions in _.rubocop.yml_  now are a won't fix?

hferentschik (commented):
Basically looking fine. Just wondering regarding the separation in the rubocop config files.

.rubocop_todo.yml
brgnepal: @hferentschik Yes agreed. Updated with the required changes.

hferentschik (approved):
Looking good. I'll run it locally and merge it all works. Much better split now in Rubocop config. Thanks.

brgnepal:
@hferentschik Sure, my other PRs are also dependent on it now.

hferentschik:
Fix for issue #412

PR #414, opened 2016-10-17
Fix #327 and #326 Tests and docs for --cli-version and --path options

### LINKED ISSUE #326: Implement the --cli-version and --path options for install-cli

### LINKED ISSUE #327: Add tests for install-cli for CDK
Atm we have not Cucumber tests which verify the install-cli behavior for CDK. This should go into a dedicated feature file, since the behavior differs from the ADB one. It would be easy to add, but add even more test time for something which is effectively not supported. 

That said, `install-cli` should preferably not need a running VM when used on CDK. Maybe tests can be added once the issue about help command needing running VM gets resolved. With a generic approach to this any command should be configurable to require a running VM or not.

brgnepal:
Fix #327 and #326
- Updated help for install-cli
- Also added checks to validate path given with --path option
- Added documentation for --cli-version and --path options

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -41,3 +41,6 @@ Metrics/AbcSize:
   Exclude:
     - 'lib/vagrant-service-manager/command.rb'
     - 'Rakefile'
+
+Lint/Eval:
+  Enabled: false
--- a/commands.adoc
+++ b/commands.adoc
@@ -223,9 +223,15 @@ Similarly, you can use `$ vagrant service-manager status openshift`,
 
 If you need to install the client binary for a specified service, use: +
 
-`vagrant service-manager install-cli [service]` +
+`vagrant service-manager install-cli [service] [options]` +
 
 The possible options for service are `docker` and `openshift`. +
+
+The following options can be used: +
+
+`--cli-version value`::  Specify the version of the binary as value to be installed +
+`--path value`::   Specify the absolute or relative path as value where the binary should be installed +
+`-h, --help`::   Prints help for the specific command being run.
 
 As per default, binaries are downloaded to
  `$VAGRANT_HOME/data/service-manager/bin/<service>/<cli-version>`, where
@@ -236,11 +242,11 @@ As per default, binaries are downloaded to
 To install client binary for Docker:
 
 ----------------------------------------------------------------------------------------------------
-$ vagrant service-manager install-cli docker
-# Binary already available at /home/johndoe/.vagrant.d/data/service-manager/bin/docker/1.9.1/docker
+$ vagrant service-manager install-cli docker --cli-version 1.9.1 --path /home/johndoe/bin/docker
+# Binary already available at /home/johndoe/bin/docker
 # run binary as:
 # docker <command>
-export PATH=/home/johndoe/.vagrant.d/data/service-manager/bin/docker/1.9.1:$PATH
+export PATH=/home/johndoe/bin:$PATH
 
 # run following command to configure your shell:
 # eval "$(VAGRANT_NO_COLOR=1 vagrant service-manager install-cli docker | tr -d '\r')"
--- a/features/adb-openshift.feature
+++ b/features/adb-openshift.feature
@@ -57,6 +57,14 @@ Feature: Command output from various OpenShift related commands in ADB
     Then the exit status should be 0
     And the binary "oc" should be installed
 
+    When I run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0`
+    Then the exit status should be 0
+    And the binary "oc" of service "openshift" should be installed with version "1.3.0"
+
+    When I evaluate and run `bundle exec vagrant service-manager install-cli openshift --path #{ENV['VAGRANT_HOME']}/oc`
+    Then the exit status should be 0
+    And the binary should be installed in path "#{ENV['VAGRANT_HOME']}/oc"
+
     When I successfully run `bundle exec vagrant reload`
     And I successfully run `bundle exec vagrant service-manager status openshift`
     Then the exit status should be 0
--- a/features/cdk-openshift.feature
+++ b/features/cdk-openshift.feature
@@ -70,6 +70,19 @@ Feature: Command output from various OpenShift related commands in CDK
     Then the exit status should be 0
     And the binary "oc" should be installed
 
+    When I run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0`
+    Then the exit status should be 0
+    And the binary "oc" of service "openshift" should be installed with version "1.3.0"
+
+    When I evaluate and run `bundle exec vagrant service-manager install-cli openshift --path #{ENV['VAGRANT_HOME']}/oc`
+    Then the exit status should be 0
+    And the binary should be installed in path "#{ENV['VAGRANT_HOME']}/oc"
+
+    When I successfully run `bundle exec vagrant reload`
+    And I successfully run `bundle exec vagrant service-manager status openshift`
+    Then the exit status should be 0
+    And the service "openshift" should be running
+
     Examples:
       | box   | provider   | ip          |
       | cdk   | virtualbox | 10.10.10.42 |
--- a/features/install-cli.feature
+++ b/features/install-cli.feature
@@ -35,7 +35,9 @@ Feature: Command behavior of client side tools installation
       A supported service. For example: docker, kubernetes or openshift.
 
     Options:
-          -h, --help         print this help
+          -h, --help            print this help
+          --cli-version         binary version to install
+          --path                absolute or relative path where to install the binary
 
     Example:
           vagrant service-manager install-cli docker
@@ -51,6 +53,14 @@ Feature: Command behavior of client side tools installation
     When I run `bundle exec vagrant service-manager install-cli docker`
     Then the exit status should be 0
     And the binary "docker" should be installed
+
+    When I run `bundle exec vagrant service-manager install-cli docker --cli-version 1.12.1`
+    Then the exit status should be 0
+    And the binary "docker" of service "docker" should be installed with version "1.12.1"
+
+    When I evaluate and run `bundle exec vagrant service-manager install-cli docker --path #{ENV['VAGRANT_HOME']}/docker`
+    Then the exit status should be 0
+    And the binary should be installed in path "#{ENV['VAGRANT_HOME']}/docker"
 
     Examples:
       | box   | provider   | ip          |
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -150,5 +150,25 @@ Then(/^the binary "([^"]*)" should be installed$/) do |binary|
   end
 
   expect(binaries.size).to eq(1)
-  expect(File.executable?(binaries[0])).to eq(true)
+  expect(File.executable?(binaries.first)).to eq(true)
+end
+
+Then(/^the binary "([^"]*)" of service "([^"]*)" should be installed with version "([^"]*)"$/) do |b, s, v|
+  binary_path = "#{ENV['VAGRANT_HOME']}/data/service-manager/bin/#{s}/#{v}/#{b}"
+  expect(File.file?(binary_path)).to eq(true)
+  expect(File.executable?(binary_path)).to eq(true)
+end
+
+Then(/^the binary should be installed in path "([^"]*)"$/) do |path|
+  binary_path = eval('"' + path + '"')
+  expect(File.file?(binary_path)).to eq(true)
+  expect(File.executable?(binary_path)).to eq(true)
+end
+
+# Per default the built-in Aruba 'When I run' will not evaluate/interpolate the string.
+# We create a custom step definition to work around this
+When(/^I evaluate and run `([^`]*)`$/) do |cmd|
+  cmd = eval('"' + cmd + '"')
+  cmd = sanitize_text(cmd)
+  run_simple(cmd, fail_on_error: false)
 end
--- a/lib/vagrant-service-manager/plugin_util.rb
+++ b/lib/vagrant-service-manager/plugin_util.rb
@@ -160,9 +160,11 @@ module VagrantPlugins
         path_string = ENV['PATH']
 
         separator = ':' unless path_string.include?(';')
-        path_string.split(separator).detect do |dir_path|
+        oc_path_dir = path_string.split(separator).detect do |dir_path|
           File.exist? "#{dir_path}\\oc.exe"
         end
+
+        oc_path_dir.nil? ? nil : oc_path_dir + '\oc.exe'
       end
     end
   end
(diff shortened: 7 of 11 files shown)


brgnepal:
@hferentschik Lets get rid of this PR :smile:

features/install-cli.feature
```diff
@@ -52,7 +54,18 @@ Feature: Command behavior of client side tools installation
...
+    # '../vagrant.d' maps to ENV['VAGRANT_HOME']
+    When I run `bundle exec vagrant service-manager install-cli docker --path ../vagrant.d/docker`
+    Then the exit status should be 0
+    And the binary should be installed in path "ENV['VAGRANT_HOME']/docker"
+
     Examples:
       | box   | provider   | ip          |
       | adb   | virtualbox | 10.10.10.42 |
       | adb   | libvirt    | 10.10.10.42 |
+      | cdk   | virtualbox | 10.10.10.42 |
```
hferentschik: Didn't we say we keep the CDK tests in _cdk-openshift_? I am trying to remember what the plan was?

features/install-cli.feature
```diff
@@ -52,7 +54,18 @@ Feature: Command behavior of client side tools installation
     Then the exit status should be 0
     And the binary "docker" should be installed
 
+    When I run `bundle exec vagrant service-manager install-cli docker --cli-version 1.12.1`
+    Then the exit status should be 0
+    And the binary "docker" should be installed with version "1.12.1"
+
+    # '../vagrant.d' maps to ENV['VAGRANT_HOME']
+    When I run `bundle exec vagrant service-manager install-cli docker --path ../vagrant.d/docker`
```
hferentschik: I am still skeptical about _../vagrant.d/docker_. Have you tried "#{ENV['VAGRANT_HOME']/docker}?

commands.adoc
```diff
@@ -236,11 +244,11 @@ As per default, binaries are downloaded to
 To install client binary for Docker:
 
 ----------------------------------------------------------------------------------------------------
-$ vagrant service-manager install-cli docker
-# Binary already available at /home/johndoe/.vagrant.d/data/service-manager/bin/docker/1.9.1/docker
+$ vagrant service-manager install-cli docker --cli-version 1.9.1 --path /home/johndoe/docker
```
hferentschik: What's about _/home/johndoe/bin/docker_ as an example dir. No one puts a binary directly into the home directory, right?

.rubocop_todo.yml
```diff
@@ -55,3 +55,12 @@ Style/Documentation:
 Style/FileName:
   Exclude:
     - 'lib/vagrant-service-manager.rb'
+
```
hferentschik, line 32: I think there should be no more Rubocop changes required in this pull request, right? Also this seems to create a merge conflict when applying to master.

locales/en.yml
```diff
@@ -104,7 +104,9 @@ en:
             A supported service. For example: docker, kubernetes or openshift.
 
           Options:
-                -h, --help         print this help
+                -h, --help            print this help
+                --cli-version         binary version to install
```
hferentschik, line 108: we are not providing any short options for this?

hferentschik (changes requested):

.rubocop_todo.yml
```diff
@@ -55,3 +55,12 @@ Style/Documentation:
...
+Metrics/BlockLength:
+   Exclude:
+     - 'Rakefile'
+     - 'test/vagrant-service-manager/*.rb'
+     - 'test/vagrant-service-manager/binary_handlers/*.rb'
+
+Lint/Eval:
+  Enabled: false
```
hferentschik, line 46: I guess this might be needed, right?

brgnepal: Yes @hferentschik

locales/en.yml
brgnepal: @hferentschik , no. Do you think it is mandatory? we have other options like --script-readable etc too which doesn't have short options

.rubocop_todo.yml
brgnepal: @hferentschik Yes

commands.adoc
brgnepal: @hferentschik Ok agreed

features/install-cli.feature
brgnepal: @hferentschik , it will not work as `#{ENV['VAGRANT_HOME']/docker}` is not know through command line. Even when passing `$VAGRANT_HOME/docker`, ruby tried to create `$VAGRANT_HOME` directory inside aruba directory.

brgnepal:
@hferentschik All tests passed :smile:

features/adb-openshift.feature
```diff
@@ -57,6 +57,12 @@ Feature: Command output from various OpenShift related commands in ADB
     Then the exit status should be 0
     And the binary "oc" should be installed
 
+    # '../vagrant.d' maps to ENV['VAGRANT_HOME']
+    When I run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0 --path ../vagrant.d/oc`
```
hferentschik: Here is what I came up with. The issue is really that Aruba does not execute the command in backticks directly and hence interpolation will take place. What's about:

```
    When I evaluate and run `bundle exec vagrant service-manager install-cli docker --path #{ENV['VAGRANT_HOME']}/docker`
    And the binary should be installed in path "#{ENV['VAGRANT_HOME']}/docker"
```

To make this work we need a new step defintion:

```
# Per default the built-in Aruba 'When I run' will not evaluate/interpolate the string.
# We create a custom step definition to work around this
When(/^I evaluate and run `([^`]*)`$/)do |cmd|
    cmd = eval('"' + cmd + '"')
   cmd = sanitize_text(cmd)
   run_simple(cmd, :fail_on_error => false)
end
```

Note, I also changed the way ' And the binary should be installed in path' is implemented:

```
Then(/^the binary should be installed in path "([^"]*)"$/) do |path|
  binary_path = eval('"' + path + '"')
  expect(File.file?(binary_path)).to eq(true)
  expect(File.executable?(binary_path)).to eq(true)
end
```

With this in place one can consistently use '#{ENV['VAGRANT_HOME']}' (or any other variable really). I think this is a "cleaner" solution. Everything is explicit. WDYT?

commands.adoc
```diff
@@ -236,11 +244,11 @@ As per default, binaries are downloaded to
 To install client binary for Docker:
 
 ----------------------------------------------------------------------------------------------------
-$ vagrant service-manager install-cli docker
-# Binary already available at /home/johndoe/.vagrant.d/data/service-manager/bin/docker/1.9.1/docker
+$ vagrant service-manager install-cli docker --cli-version 1.9.1 --path /home/johndoe/bin/docker
+# Binary already available at /home/johndoe/bin/docker
 # run binary as:
```
hferentschik, line 247: +1

locales/en.yml
hferentschik: Works for me. Just wanted to make sure we are not missing something.

hferentschik (changes requested):
Looking good. I'd suggest to make one more change to the way the features are written, but otherwise ok now.

features/adb-openshift.feature
brgnepal: @hferentschik This works great :+1: :100: 

I didn't know that :smile:  Thanks

features/cdk-openshift.feature
```diff
@@ -70,6 +70,16 @@ Feature: Command output from various OpenShift related commands in CDK
     Then the exit status should be 0
     And the binary "oc" should be installed
 
+    When I evaluate and run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0 --path #{ENV['VAGRANT_HOME']}/oc`
+    Then the exit status should be 0
+    And the binary "oc" should be installed with version "1.3.0"
+    And the binary should be installed in path "ENV['VAGRANT_HOME']/oc"
```
hferentschik: hmm, is this right? Should it not be "#{ENV['VAGRANT_HOME']}/oc"?

hferentschik: I would expect this to fail, tbh

hferentschik (changes requested):

commands.adoc
```diff
@@ -223,10 +223,18 @@ Similarly, you can use `$ vagrant service-manager status openshift`,
 
 If you need to install the client binary for a specified service, use: +
 
-`vagrant service-manager install-cli [service]` +
+`vagrant service-manager install-cli [service] [options] [arg]` +
```
hferentschik: Do you really need '[arg]' here? The value for an option is really encapsulated in [options] itself, right? I guess one can do something like '--cli-version=' to indicate that the option takes a value.

hferentschik (changes requested):

features/support/env.rb
```diff
@@ -150,5 +150,25 @@ def extract_process_id(data)
   end
 
   expect(binaries.size).to eq(1)
-  expect(File.executable?(binaries[0])).to eq(true)
+  expect(File.executable?(binaries.first)).to eq(true)
+end
+
+Then(/^the binary "([^"]*)" should be installed with version "([^"]*)"$/) do |binary, version|
+  binary_path = "#{ENV['VAGRANT_HOME']}/data/service-manager/bin/#{version}/#{binary}"
```
hferentschik: I was wrong, we cannot build the path directly. This test fails now, there is one more level after 'bin' which depicts the service. We either would need to pass the service as well to build the full path or we indeed have to do some path matching. Maybe it is more explicit to so something like with something like:
"Then(/^the binary "([^"]_)" of service "([^"]_)" should be installed with version "([^"]*)"$/)"? I just want to make sure we don't get any false positives by using a recursive scan approach.

commands.adoc
brgnepal: @hferentschik `--cli-version=` will create more confusion IMO as user will then assume they need to enter option value as `--cli-version=1.3.0` which VSM cannot parse ATM.

Or you think the options description should include the value as:

The following options can be used:

`--cli-version value`::  Specify the version of the binary to be installed
`--path value`::   Specify the absolute or relative path where the binary should be installed
`-h, --help`::   Prints help for the specific command being run.

brgnepal:
@hferentschik Pushed the new changes which include missing service name in step definition alongwith tests fix

commands.adoc
hferentschik: That latter works as well for me. ' [options] [arg]' seems wrong. OR what do you think?

brgnepal:
@hferentschik CI seems to be passing now :+1:

commands.adoc
hferentschik: As it stands it is wrong '[options] [arg]'. It implies that options and args are disconnected

features/cdk-openshift.feature
```diff
@@ -70,6 +70,19 @@ Feature: Command output from various OpenShift related commands in CDK
     Then the exit status should be 0
     And the binary "oc" should be installed
 
+    When I run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0`
+    Then the exit status should be 0
+    And the binary "oc" of service "openshift" should be installed with version "1.3.0"
```
hferentschik, line 75: 👍 I think this is better

commands.adoc
brgnepal: @hferentschik updated the PR with the changes.

brgnepal:
@hferentschik Anything remaining in this PR now? I think it is ready to merge now.

commands.adoc
```diff
@@ -223,10 +223,16 @@ Similarly, you can use `$ vagrant service-manager status openshift`,
 
 If you need to install the client binary for a specified service, use: +
 
-`vagrant service-manager install-cli [service]` +
+`vagrant service-manager install-cli [service] [options]` +
```
hferentschik, line 226: 👍

features/adb-openshift.feature
```diff
@@ -57,6 +57,14 @@ Feature: Command output from various OpenShift related commands in ADB
     Then the exit status should be 0
     And the binary "oc" should be installed
 
+    When I run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0`
+    Then the exit status should be 0
+    And the binary "oc" of service "openshift" should be installed with version "1.3.0"
```
hferentschik, line 62: 👍 

I created issue #415 to follow up on what's going on with CI and why we did not get test failures when some of the tests were clearly wrong. We need to get to the bottom of this, otherwise we cannot trust CI results, in particular since these tests are not run per default on the dev machine

features/adb-openshift.feature
```diff
@@ -57,6 +57,14 @@ Feature: Command output from various OpenShift related commands in ADB
...
+
+    When I evaluate and run `bundle exec vagrant service-manager install-cli openshift --path #{ENV['VAGRANT_HOME']}/oc`
```
hferentschik, line 64: 👍 I like how the tests look like now. Also nice to use the same path in the _When_ and _Then_ clause and that it gets transparently interpolated.

features/install-cli.feature
```diff
@@ -52,6 +54,14 @@ Feature: Command behavior of client side tools installation
     Then the exit status should be 0
     And the binary "docker" should be installed
 
+    When I run `bundle exec vagrant service-manager install-cli docker --cli-version 1.12.1`
+    Then the exit status should be 0
+    And the binary "docker" of service "docker" should be installed with version "1.12.1"
```
hferentschik, line 59: I think this step is also quite nice now. Allows us to do a dedicated check in its implementation

lib/vagrant-service-manager/plugin_util.rb
```diff
@@ -160,9 +160,11 @@ def self.fetch_existing_oc_binary_path_in_windows
         path_string = ENV['PATH']
 
         separator = ':' unless path_string.include?(';')
-        path_string.split(separator).detect do |dir_path|
+        oc_path_dir = path_string.split(separator).detect do |dir_path|
           File.exist? "#{dir_path}\\oc.exe"
         end
+
+        oc_path_dir.nil? ? nil : oc_path_dir + '\oc.exe'
```
hferentschik, line 167: We let this sneak in for now ;-)

hferentschik (approved):

lib/vagrant-service-manager/plugin_util.rb
brgnepal: @hferentschik I had already explained it :smile: 

So, in windows, devsuite download the oc binary and place it in `C:\DevSuite\cdk\bin\oc.exe` which is also added in environment variable $PATH.

Previously, I was only checking through the directory paths presents in $PATH and checking the existence of dir_path + '\oc.exe' file. On first detection, I return the dir_path only. However, I need the actual full path not `dir_path`. 
Now, sometime if  detect doesn't match anything then it returns nil. 

Probably, I should have changed this line to 

```
oc_path_dir + '\oc.exe' unless oc_path_dir.nil?
```

hferentschik:
Yeah :-)

Issue #415, opened 2016-10-17
Investigate the behaviour of Openshift features on CI


hferentschik:
It seems it does not fail in cases where the tests are wrong, eg _When I evaluate and run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0 --path #{ENV['VAGRANT_HOME']}/oc`_ should fail, but did not.

We can temporarily change the CI job to build one of our forks on which we can introduce some obvious test errors. We need to verify that this will result in test failures.

brgnepal:
@hferentschik  I have updated the test to successfully run 

```
bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0 --path #{ENV['VAGRANT_HOME']}/oc
```

as I found it is related to minimum memory requirement for Openshift #420 .

However, there is one test case which still need to be investigated.

```
When I run `bundle exec vagrant service-manager install-cli openshift`
```

We can keep this issue open till we found the reason of the above test case failure. 
Rest is fine and genuinely working in CI as part of my investigation.

brgnepal:
We also found that there is some wait time (~20s in dev machine) required in between `vagrant up` and first operation related to openshift like `vagrant service-manager status openshift`.

We need to investigate the root cause for this delay requirement.

hferentschik:
Even the CDK OpenShift test fails after upgrading to the latest CDK box.

hferentschik:
For the record, the initial problem was that the tests were not configured to run against CDK. Hence, it looked like they were running and passing, but in reality they got skipped. This was hidden by issue #419. We were using the default pretty formatter without coloring which did not indicate which tests got run and which test got skipped.

hferentschik:
It seems there is a regression in the OpenShift service startup which leads to the fact that the OpenShift status is not directly "running" after a successful 'vagrant up'.

hferentschik:
So using CDK 2.2, the Vagrant configuration from adb-atomic-developer-bundle and running:

```
vagrant up; vagrant service-manager status
```

I get:

```
==> default: Docker service configured successfully...
==> default: OpenShift service configured successfully...
==> default: Mounting SSHFS shared folder...
==> default: Mounting folder via SSHFS: /Users/hardy => /Users/hardy
==> default: Checking Mount..
==> default: Folder Successfully Mounted!
==> default: Running provisioner: shell...
    default: Running: inline script
==> default: Running provisioner: shell...
    default: Running: inline script
==> default:
[...]

PR #416, opened 2016-10-19
Fix #376 --path option of install cli not working as expected

### LINKED ISSUE #376: --path feature for install cli not working as expected
--path feature for install cli not working as expected
Find issue details : https://gist.github.com/naina-verma/e03bdfb7bc60adaa22b2dbb75e48f604
tested on latest vsm 
`vagrant-registration (1.2.3)
  Version Constraint: 1.2.3
vagrant-service-manager (1.3.0.dev)
   Version Constraint: 1.3.0.dev
vagrant-sshfs (1.2.0)`

brgnepal:
Fix #376
- install-cli command will fail if invalid --path is given or
  path needs permission

### DIFF
--- a/features/adb-openshift.feature
+++ b/features/adb-openshift.feature
@@ -19,6 +19,9 @@ Feature: Command output from various OpenShift related commands in ADB
       config.vm.network :private_network, ip: '<ip>'
       config.vm.synced_folder '.', '/vagrant', disabled: true
 
+      config.vm.provider('libvirt') { |v| v.memory = 3072 }
+      config.vm.provider('virtualbox') { |v| v.memory = 3072 }
+
       config.servicemanager.services = 'docker, openshift'
     end
     """
@@ -30,7 +33,12 @@ Feature: Command output from various OpenShift related commands in ADB
     ==> default: OpenShift service configured successfully...
     """
 
-    When I successfully run `bundle exec vagrant service-manager env openshift`
+    When I sleep for 10 seconds
+    And I successfully run `bundle exec vagrant service-manager status`
+    Then stdout from "bundle exec vagrant service-manager status" should contain "openshift - running"
+
+    When I sleep for 10 seconds
+    And I successfully run `bundle exec vagrant service-manager env openshift`
     Then stdout from "bundle exec vagrant service-manager env openshift" should be evaluable in a shell
     And stdout from "bundle exec vagrant service-manager env openshift" should contain:
     """
@@ -53,23 +61,25 @@ Feature: Command output from various OpenShift related commands in ADB
     DOCKER_REGISTRY=hub.openshift.centos7-adb.<ip>.xip.io
     """
 
-    When I run `bundle exec vagrant service-manager install-cli openshift`
-    Then the exit status should be 0
-    And the binary "oc" should be installed
-
-    When I run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0`
+    When I sleep for 10 seconds
+    And I successfully run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0`
     Then the exit status should be 0
     And the binary "oc" of service "openshift" should be installed with version "1.3.0"
+    And stdout from "bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0" should be evaluable in a shell
 
-    When I evaluate and run `bundle exec vagrant service-manager install-cli openshift --path #{ENV['VAGRANT_HOME']}/oc`
+    When I sleep for 10 seconds
+    And I evaluate and run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0 --path #{ENV['VAGRANT_HOME']}/oc`
     Then the exit status should be 0
     And the binary should be installed in path "#{ENV['VAGRANT_HOME']}/oc"
+    And stdout after evaluating and running "bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0 --path #{ENV['VAGRANT_HOME']}/oc" should be evaluable in a shell
 
     When I successfully run `bundle exec vagrant reload`
+    And I sleep for 10 seconds
     And I successfully run `bundle exec vagrant service-manager status openshift`
     Then the exit status should be 0
     And the service "openshift" should be running
 
     Examples:
       | box   | provider   | ip          |
+      | adb   | libvirt    | 10.10.10.42 |
       | adb   | virtualbox | 10.10.10.42 |
--- a/features/cdk-openshift.feature
+++ b/features/cdk-openshift.feature
@@ -18,13 +18,25 @@ Feature: Command output from various OpenShift related commands in CDK
       config.vm.box_url = 'file://../../.boxes/<box>-<provider>.box'
       config.vm.network :private_network, ip: '<ip>'
       config.vm.synced_folder '.', '/vagrant', disabled: true
+
+      config.vm.provider('libvirt') { |v| v.memory = 3072 }
+      config.vm.provider('virtualbox') { |v| v.memory = 3072 }
+
       config.servicemanager.services = 'docker, openshift'
     end
     """
 
     When I successfully run `bundle exec vagrant up --provider <provider>`
-    # TODO, for some reason I can not use 'successfully' here. Seems the exit code is not 0 in this case!?
-    And I run `bundle exec vagrant service-manager env openshift`
+    Then stdout from "bundle exec vagrant up --provider <provider>" should contain:
+    """
+    ==> default: Docker service configured successfully...
+    ==> default: OpenShift service configured successfully...
+    """
+
+    When I successfully run `bundle exec vagrant service-manager status`
+    Then stdout from "bundle exec vagrant service-manager status" should contain "openshift - running"
+
+    When I successfully run `bundle exec vagrant service-manager env openshift`
     Then stdout from "bundle exec vagrant service-manager env openshift" should be evaluable in a shell
     And stdout from "bundle exec vagrant service-manager env openshift" should contain:
     """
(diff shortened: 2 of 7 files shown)


brgnepal:
@hferentschik CI is passing :+1:

features/install-cli.feature
```diff
@@ -62,6 +62,14 @@ Feature: Command behavior of client side tools installation
     Then the exit status should be 0
     And the binary should be installed in path "#{ENV['VAGRANT_HOME']}/docker"
 
+    When I evaluate and run `bundle exec vagrant service-manager install-cli docker --path #{ENV['VAGRANT_HOME']}/unknown_dir/docker`
+    Then the exit status should be 126
+    And stderr from evaluating and running "bundle exec vagrant service-manager install-cli docker --path #{ENV['VAGRANT_HOME']}/unknown_dir/docker" should match /Directory path #{ENV['VAGRANT_HOME']}/unknown_dir is invalid or doesn't exist/
```
hferentschik, line 70: Nice 👍

hferentschik: What's about testing/verifying that the error message in the original bug report is not in the output? Neither here, nor in any of the successful cases of install-cli

brgnepal:
retest this please

brgnepal:
@hferentschik The tests seems to be passing now. This time these are passing genuinely :smile:. However, there is one testcase which is failing in terms of openshift
`bundle exec vagrant service-manager install-cli openshift` which I don't know. I tried a lot for its debugging.

However, its cli-version `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.1` is passing.
This PR has --cli-version.

The failed one is here https://ci.centos.org/job/vagrant-service-manager-budh/11/console.

However, no such issue in case of `cdk-openshift.feature` (https://github.com/budhrg/vagrant-service-manager/blob/e81d07a3e4b73042a1f633391318696e6e7750f0/features/cdk-openshift.feature).

I believe if you have not any other review, we can merge it and investigate the edge case in this issue https://github.com/projectatomic/vagrant-service-manager/issues/415

Let me know what is your opinion.

brgnepal:
retest this please

features/adb-openshift.feature
```diff
@@ -30,6 +34,11 @@ Feature: Command output from various OpenShift related commands in ADB
     ==> default: OpenShift service configured successfully...
     """
 
+    When I successfully run `sleep 20`
```
hferentschik: That's a pretty long timeout. Really imo this also an issue of the underlying service setup. I expect that when Vagrant up has completed, everything is up and running, but you seem to imply this is not the case.

features/adb-openshift.feature
```diff
@@ -30,6 +34,11 @@ Feature: Command output from various OpenShift related commands in ADB
...
+    When I successfully run `bundle exec vagrant service-manager status`
```
hferentschik: "And I sucessfully" ...

features/adb-openshift.feature
```diff
@@ -30,6 +34,11 @@ Feature: Command output from various OpenShift related commands in ADB
...
+    Then stdout from "bundle exec vagrant service-manager status" should contain "openshift - running"
+
+    When I successfully run `sleep 20`
```
hferentschik: Do you have an idea why this is necessary?

hferentschik: Also, I think we are better of to create our own step definition. Something like "When I sleep 20 seconds". Then in the implementation you are calling Ruby's sleep method. This way you are doing things within Ruby. Using `sleep 2-` effectively calls the shells sleep command which might or might not work on Windows.

features/adb-openshift.feature
```diff
@@ -53,23 +62,23 @@ Feature: Command output from various OpenShift related commands in ADB
...
+    When I successfully run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0`
     Then the exit status should be 0
     And the binary "oc" of service "openshift" should be installed with version "1.3.0"
+    And stdout from "bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0" should be evaluable in a shell
 
-    When I evaluate and run `bundle exec vagrant service-manager install-cli openshift --path #{ENV['VAGRANT_HOME']}/oc`
+    When I evaluate and run `bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0 --path #{ENV['VAGRANT_HOME']}/oc`
     Then the exit status should be 0
     And the binary should be installed in path "#{ENV['VAGRANT_HOME']}/oc"
+    And stdout from "bundle exec vagrant service-manager install-cli openshift --cli-version 1.3.0 --path #{ENV['VAGRANT_HOME']}/oc" should be evaluable in a shell
```
hferentschik: 👍

features/cdk-openshift.feature
```diff
@@ -18,11 +18,23 @@ Feature: Command output from various OpenShift related commands in CDK
...
+      end
+
       config.servicemanager.services = 'docker, openshift'
     end
     """
 
     When I successfully run `bundle exec vagrant up --provider <provider>`
+    Then stdout from "bundle exec vagrant up --provider <provider>" should contain:
+    """
+    ==> default: Docker service configured successfully...
```
hferentschik, line 32: 👍

features/install-cli.feature
```diff
@@ -53,14 +53,25 @@ Feature: Command behavior of client side tools installation
...
+    And the stderr should not contain anything
 
     When I evaluate and run `bundle exec vagrant service-manager install-cli docker --path #{ENV['VAGRANT_HOME']}/docker`
     Then the exit status should be 0
     And the binary should be installed in path "#{ENV['VAGRANT_HOME']}/docker"
+    And the stderr should not contain anything
+
+    When I evaluate and run `bundle exec vagrant service-manager install-cli docker --path #{ENV['VAGRANT_HOME']}/unknown_dir/docker`
+    Then the exit status should be 126
+    And stderr from evaluating and running "bundle exec vagrant service-manager install-cli docker --path #{ENV['VAGRANT_HOME']}/unknown_dir/docker" should match /Directory path #{ENV['VAGRANT_HOME']}/unknown_dir is invalid or doesn't exist/
```
hferentschik, line 70: 👍

features/support/env.rb
```diff
@@ -93,6 +93,12 @@ def extract_process_id(data)
   aruba.command_monitor.find(Aruba.platform.detect_ruby(cmd)).send(:stdout) =~ /#{regexp}/
 end
 
+Then(%r{^stderr from evaluating and running "([^"]*)" should match /(.*)/$}) do |cmd, regexp|
```
hferentschik, line 101: :-)

lib/vagrant-service-manager/installer.rb
```diff
@@ -23,21 +24,30 @@ def install
           @binary_handler.install
         end
 
-        @binary_handler.print_message
+        @binary_handler.print_message(@option_msg)
+      end
+
+      def format_options
+        msg = ''
```
hferentschik, line 31: Is this to handle the right output for the 'eval'?

hferentschik (changes requested):
Getting there. Looking quite good now. Not quite sure what to think about 20 seconds timeout. Does it really have to be so high. We also needs to get to the bottom of this from a CDK side of things. There should be no wait required really. 

For me to get through the test suite, I still need https://github.com/hferentschik/vagrant-service-manager/commit/19cf82481a843d9f4e16ad6421bcb3dde1132028

hferentschik:
retest

hferentschik:
retest

hferentschik:
test this please

lib/vagrant-service-manager/installer.rb
brgnepal: @hferentschik Yes

brgnepal:
@hferentschik Hi, I have updated the tests now. It include both evalauable the install-cli command and added step for sleep too.

brgnepal:
retest this please

hferentschik:
test this please

features/adb-openshift.feature
```diff
@@ -19,6 +19,10 @@ Feature: Command output from various OpenShift related commands in ADB
       config.vm.network :private_network, ip: '<ip>'
       config.vm.synced_folder '.', '/vagrant', disabled: true
 
+      config.vm.provider "libvirt" do |v, _override|
+        v.memory = 3072
+      end
+
```
hferentschik, line 24: Why don't you add the virtualbox memory settings here as well?

hferentschik:
@budhrg Did you create an issue somewhere in adb to find out why we need a wait period before between 'vagrant up' and the service-manager 'status'? That does not seem right to me. 

Also, lets create a follow up issue in service-manager to find a more direct condition than sleeping 10 seconds. This might or might not work depending on the machine the tests are running. It could be something like "And wait max x seconds to OpenShift to be operational" where the implementation checks in a loop some condition of the OpenShift cluster and times out after max seconds. Adding random waits to a test is a bad smell.

hferentschik:
@budhrg - does this `bundle exec rake clean features BOX=adb,cdk` pass all tests on your work station?

hferentschik:
@budhrg Didn't we want to enable the Kubernates features on CI as well?

hferentschik (changes requested):
The tests are not passing! You need to convert the captured string into an int to make the sleep work.

```
      can't convert String into time interval (TypeError)
  /root/vagrant-service-manager/feature/support/env.rb:188:in `sleep'
```

hferentschik:
@budhrg For what it's worth `bundle exec rake clean features BOX=adb,cdk` seems to work for me now ignoring the string conversion error (note this in itself makes of course all adb openshift tests to be skipped)

features/adb-openshift.feature
```diff
@@ -30,7 +38,12 @@ Feature: Command output from various OpenShift related commands in ADB
     ==> default: OpenShift service configured successfully...
     """
 
-    When I successfully run `bundle exec vagrant service-manager env openshift`
+    When I sleep for 10 seconds
+    And I successfully run `bundle exec vagrant service-manager status`
+    Then stdout from "bundle exec vagrant service-manager status" should contain "openshift - running"
+
+    When I sleep for 10 seconds
```
LalatenduMohanty, line 40: This should not be the case. We need to investigate the root cause of why we need 10s delay.

hferentschik (approved):
:+1

PR #431, opened 2016-10-27
Fix #335 'bundle exec vagrant up --provider libvirt' should work

### LINKED ISSUE #335: "bundle exec vagrant up --provider=libvirt" should work
This needs a change in Gemspec file i.e. we need to add `gem "vagrant-libvirt"` to the `group :plugins` section. Refer [1] for details. 
[1] https://github.com/vagrant-libvirt/vagrant-libvirt/issues/563

brgnepal:
Fix #335 

We can now able to do `bundle exec vagrant up --provider libvirt`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,8 +4,6 @@ group :development do
   gem 'vagrant',
       git: 'https://github.com/mitchellh/vagrant.git',
       ref: 'v1.8.4'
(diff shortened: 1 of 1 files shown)


hferentschik (approved):
Looking good to me. See whether CI passes which runs libvirt as well and I double check that nothing breaks on my OS X

Issue #389, opened 2016-09-26
Support the use of install-cli in HTTP proxy environment


LalatenduMohanty:
**OS details:**
Windows 7

**Provider:**
VirtualBox

**Output of `vagrant plugin list`:**

```
vagrant-registration (1.3.0)
vagrant-service-manager (1.3.3)
```

**Steps to reproduce the issue:**
1.  A Windows proxy environment. Only web browser has access to internet with proxy.
2. vagrant service-manager install-cli openshift

`vagrant service-manager install-cli openshift` should download and install oc binary. At this point the workstation's CMD or Cygwin does not have access to internet through proxy.  So VSM should setup the proxy for the Vagrant environment and then use it to download stuff for `install-cli` feature.

hferentschik:
>  Only web browser has access to internet with proxy.

Is this really what you mean? IMO a HTTP proxy environment means that all HTTP related traffic needs to go through a proxy server. This does not mean though that this traffic must originate from the browser. Any application can open the relevant ports and create an HTTP connection. So in the case of service-manager, it should be as simple as using the proxy credentials which can be configured already and applying them to the code which establishes the HTTP connection. 

I am not quite sure how CMD or Cygwin bash come into play here. They themselves are not making connections, right? Any tool launched via CMD or cygwin shell though would need to use the right proxy settings.

hferentschik:
>  At this point the workstation's CMD or Cygwin does not have access to internet through proxy. 

What's the error message?

LalatenduMohanty:
@hferentschik In Windows you can not set proxy settings for the whole system. The proxy server information needs to be set per application. Typically the web browser is configured to use the proxy server and user can access internet. However the Vagrant environment does not have proxy server information. VSM downloads the binaries using install-cli command which is not aware of proxy server information. 

Regarding the the CMD and Cygwin , if you are running any command in them to access internet the environment needs to be aware of proxy server information to route the request through proxy server.

LalatenduMohanty:
> So in the case of service-manager, it should be as simple as using the proxy credentials which can be configured already and applying them to the code which establishes the HTTP connection.

Yes, thats right. That's what I mean when I say VSM should support proxy.  I think you disagree that if configure the shell environment to use proxy and then can run VSM commands there and it should work.

hferentschik:
> I think you disagree that if configure the shell environment to use proxy and then can run VSM commands there and it should work. 

[...]

PR #435, opened 2016-10-28
Fix #434 Align proxy config names similar to 'sccli'

### LINKED ISSUE #434: Align proxy config names similar to 'sccli'
Currently, proxy configurations names are `http_proxy`, `http_proxy_user` and `http_proxy_password` which seems to specific to `http` and may confuse user to think about different configurations for `https` proxy.

Let's use config names as `proxy`, `proxy_user` and `proxy_password` as used by [sccli here](https://github.com/projectatomic/adb-utils/blob/master/utils/sccli.py#L70-L76).

Also, proposing this change as I found bug in setting proxy through service-manager that proxy is not being setup properly.

Command being executed through service-manager is 

```
HTTP_PROXY='url' HTTP_PROXY_USER='username' HTTP_PROXY_PASSWORD='password' sccli openshift
```

which is a wrong command.
Actual command should be used is: 

```
PROXY='url' PROXY_USER='username' PROXY_PASSWORD='password' sccli openshift
```

brgnepal:
Fix #434

Now, I can see the proxy related changes inside VM.

```
$ bundle exec vagrant ssh
Last login: Fri Oct 28 12:47:16 2016
[vagrant@rhel-cdk ~]$ cat /etc/sysconfig/openshift_option 
# /etc/sysconfig/openshift_options

# Modify these options if you want to change openshift hostname
OPENSHIFT_SUBDOMAIN="$(hostname).$(/opt/adb/openshift/get_ip_address).xip.io"
NO_PROXY=127.0.0.1,10.0.2.15,10.10.10.42,172.17.0.1,.xip.io,172.30.0.0/16,172.17.0.0/16,rhel-cdk
HTTP_PROXY=http://fedora:fedora123@http://10.70.49.109:3128
HTTPS_PROXY=http://fedora:fedora123@http://10.70.49.109:3128
IMAGE="registry.access.redhat.com/openshift3/ose:v3.2.1.9"

[vagrant@rhel-cdk ~]$ cat /etc/systemd/system/docker.service.d/http-proxy.conf
[Service]
Environment="HTTP_PROXY=http://fedora:fedora123@http://10.70.49.109:3128" "NO_PROXY=localhost,127.0.0.1,::1,.xip.io,172.30.66.245"
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -153,18 +153,20 @@ have IPv4 addresses, the address on eth2 is used.
 === HTTP Proxy Settings
 
 In an environment where the HTTP traffic needs to pass through an
-HTTP proxy server, set the `http_proxy`, `http_proxy_user` and
-`http_proxy_password` proxy configurations in the Vagrantfiles to enable
+HTTP proxy server, set the `proxy`, `proxy_user` and
+`proxy_password` proxy configurations in the Vagrantfiles to enable
 services like Docker and OpenShift to function correctly. +
 You can do so via:
  +
 -----
-config.servicemanager.http_proxy = <Proxy URL>
-config.servicemanager.http_proxy_user = <Proxy user name>
-config.servicemanager.http_proxy_password = <Proxy user password>
+config.servicemanager.proxy = <Proxy URL>
+config.servicemanager.proxy_user = <Proxy user name>
+config.servicemanager.proxy_password = <Proxy user password>
 -----
 
-When these settings are applied, they are passed through to the Docker and OpenShift services. In an unauthenticated proxy environment, the `http_proxy_user` and `http_proxy_password` configurations can be omitted.
+When these settings are applied, they are passed through to the Docker and OpenShift services. In an unauthenticated proxy environment, the `proxy_user` and `proxy_password` configurations can be omitted.
+
+NOTE: `http_proxy`, `http_proxy_user` and `http_proxy_password` have been depreacted now.
 
 === Configuring Services
 The vagrant-service-manager helps you configure the service of your choice:
--- a/lib/vagrant-service-manager/config.rb
+++ b/lib/vagrant-service-manager/config.rb
@@ -7,7 +7,7 @@ module VagrantPlugins
     OPENSHIFT_CONFIG = [
       :openshift_docker_registry, :openshift_image_name, :openshift_image_tag
     ].freeze
-    PROXY_CONFIG = [:http_proxy, :http_proxy_user, :http_proxy_password].freeze
+    PROXY_CONFIG = [:proxy, :proxy_user, :proxy_password].freeze
 
     class Config < Vagrant.plugin('2', :config)
       attr_accessor(*(BASE_CONFIG + OPENSHIFT_CONFIG + PROXY_CONFIG))
--- a/lib/vagrant-service-manager/service_base.rb
+++ b/lib/vagrant-service-manager/service_base.rb
@@ -21,7 +21,7 @@ module VagrantPlugins
       def proxy_cmd_options
         options = ''
 
-        return options unless http_proxy_settings_valid?
+        return options unless proxy_settings_valid?
 
         PROXY_CONFIG.each do |key|
           options += "#{key.to_s.upcase}='#{@machine.config.servicemanager.send(key)}' "
@@ -30,10 +30,10 @@ module VagrantPlugins
         options.chop
       end
 
-      def http_proxy_settings_valid?
-        proxy = @machine.config.servicemanager.send('http_proxy')
-        user = @machine.config.servicemanager.send('http_proxy_user')
-        password = @machine.config.servicemanager.send('http_proxy_password')
+      def proxy_settings_valid?
+        proxy = @machine.config.servicemanager.send('proxy')
+        user = @machine.config.servicemanager.send('proxy_user')
+        password = @machine.config.servicemanager.send('proxy_password')
 
         if proxy && user.nil? && password.nil? || (proxy && user && password)
           PluginLogger.debug('Detected proxy settings. Going to apply them to service commands.')
--- a/test/vagrant-service-manager/service_base_test.rb
+++ b/test/vagrant-service-manager/service_base_test.rb
@@ -8,59 +8,59 @@ module VagrantPlugins
     let(:instance) { ServiceManager::ServiceBase.new(machine, machine.env) }
 
     it 'should build the proxy options if only proxy url is given' do
-      machine.config.servicemanager.http_proxy = 'foo:8080'
+      machine.config.servicemanager.proxy = 'foo:8080'
       options = instance.proxy_cmd_options
-      assert_match(/HTTP_PROXY='foo:8080'/, options)
+      assert_match(/PROXY='foo:8080'/, options)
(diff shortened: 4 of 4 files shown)


README.adoc
```diff
@@ -157,18 +157,18 @@ have IPv4 addresses, the address on eth2 is used.
 === HTTP Proxy Settings
 
 In an environment where the HTTP traffic needs to pass through an
-HTTP proxy server, set the `http_proxy`, `http_proxy_user` and
-`http_proxy_password` proxy configurations in the Vagrantfiles to enable
+HTTP proxy server, set the `proxy`, `proxy_user` and
```
hferentschik, line 156: I don't think we can just change at will here. There was already a release with the _http.*_ variables. If we just change the names now we potentially existing users. So to do this properly, one needs to "deprecate" the old setting and for example print a warning when it is used, but still accept it. 

One also need to thing about introducing something like a migration guide which would document these type of changes between releases.

hferentschik (changes requested):
I think the easier option is to leave the config option and just make sure that the right settings are applied when making the sccli call. And add tests. TBH, if you want to make this change now, I'd also like to see a test.

README.adoc
hferentschik: One thing one could argue is that even though released, the proxy settings did not work anyways. 

@LalatenduMohanty WDYT?

brgnepal: @hferentschik @LalatenduMohanty Yes, it seems that the released proxy feature is not working due to passing wrong environment variables `HTTP_PROXY` etc instead of `PROXY` to sccli.

IMO, it is ok to release this without any depreciation since most of the users are using shipped Vagrantfiles which has properly set proxy configurations.
Otherwise we would have got at least one mail of proxy not being setup properly if they were using 
service-manager way.

WDYT ?

LalatenduMohanty: > it seems that the released proxy feature is not working due to passing wrong environment variables HTTP_PROXY etc instead of PROXY to sccli.

@budhrg How did it pass our testing of the PR?

LalatenduMohanty: Ideally we should not change user facing variables of config names and communicate if we do something like that as @hferentschik  mentioned. However in this case we should be ok to change it as it was not working at all before and we are trying the fix the issue. Also the support was added last release , so we should fix it asap before users start using it. That means we should mark this issue as a blocker issue and do a bug fix release asap with the fix.

LalatenduMohanty: > One also need to thing about introducing something like a migration guide which would document these type of changes between releases.

@hferentschik Agree. Basically it should contain all changes that are not backward compatible.

LalatenduMohanty: @budhrg @hferentschik after thinking little bit more, I think we should mention deprecation of `http_` prefix in the documentation [1] in this PR and also update [2]

[1] https://github.com/projectatomic/vagrant-service-manager#http-proxy-settings
[2] https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/installing.adoc#step-6-optional-set-up-adb-behind-an-http-proxy

LalatenduMohanty (changes requested):

README.adoc
brgnepal: @LalatenduMohanty @hferentschik then in this case we can hold this PR https://github.com/projectatomic/adb-atomic-developer-bundle/pull/572 for the time until this PR is merged and update the PR

hferentschik: > How did it pass our testing of the PR? 

That's something I would like to know as well. Obviously this is not so easy to test, but that makes it even more important to also do manual testing. 

I wonder whether we can add some cucumber tests for this. We could start a proxy as part of the test setup, but it needs to bind to an address which is then reachable from the VM. Should be possible.

@LalatenduMohanty So you prefer the variables to not have the _http_ prefix?

hferentschik: @LalatenduMohanty ^^

LalatenduMohanty: @hferentschik I do not have a strong opinion for either of these names. We should stick to one that make sense and it should work.

LalatenduMohanty: @hferentschik It seems vagrant-registration does not use `http` prefix , so may be we should get rid of it and be consistent.

hferentschik: >  It seems vagrant-registration does not use http prefix , so may be we should get rid of it and be consistent. 

Now this is a good point. Makes me swing towards removing it as well :-)
Remaining question, do we want to invest into testing this, or are we ok merging in good faith. 

@budhrg did you manually tests and verify that this will work? 
@budhrg did you already think about or even start with some acceptance test? (if you got the testing stuff we discussed for install-cli via proxy, it should not be so hard to adopt this for feature testing)

brgnepal: @hferentschik I manually tested it and its working now. Previously, it wasn't.

I have some idea about the test using rspec which we discussed for `install-cli` proxy feature.

I idea is to run HTTPServer as forward proxy server to a dummy simple http server. Then, pass details and check whether it passes through forward proxy server or not. Similar to what we discussed.

Note, that I am using a dummy simple http server as I couldn't find a way for proxy server to always return some fixed content.

brgnepal: @hferentschik , I was thinking to add the proxy test here but I think we cannot do so for this change.

The implementation we discussed during `install-cli  proxy` feature is different than this feature. By this I mean, here we have only config related changes. The implementation we have discussed need the http requests to go via proxy server and get verified.  

I will add this implementation as part of fix #389 

I have manually tested this PR and its working fine. 
One can try this patched gem from here https://www.dropbox.com/s/59w94p8slsllkbb/vagrant-service-manager-1.5.0.dev.gem?dl=0.

Let me know what do you think?

hferentschik: >  I manually tested it and its working now. Previously, it wasn't.

I take your word for it

> I idea is to run HTTPServer as forward proxy server to a dummy simple http server. 

HTTPServer? How come. I don't understand what you are planning to do. WEBrick::HTTPProxyServer should do imo, but I might be missing a piece.

> Note, that I am using a dummy simple http server as I couldn't find a way for proxy server to always return some fixed content.

Override the service method. AFAICT it contains the request and response instances. Using the response instance you should be able to return whatever you want.

> I was thinking to add the proxy test here but I think we cannot do so for this change.

I don't quite understand why not, but whatever ...

> The implementation we discussed during install-cli proxy feature is different than this feature.

Slightly different sure, but only varies in a small degree. If one can do the discussed test for _install-cli_ I am pretty sure one can test the proxy config settings as well.

OOI what have you tried? Do you have any code? Any particular problem or stack trace?

hferentschik (commented):
@budhrg - I think you have not addressed this commend from @LalatenduMohanty yet: "after thinking little bit more, I think we should mention deprecation of http_ ". @LalatenduMohanty do you still want this? I don't mind either way.

README.adoc
brgnepal: @hferentschik https://gist.github.com/budhrg/00a447d75d16809ec5f50a98543c35ca.

It is even passing that first dummy test case of checking http server is running fine or not.

LalatenduMohanty (approved):

LalatenduMohanty:
LGTM.

PR #437, opened 2016-11-04
Fix #362 Adding OpenShift configuration options into VSM Readme

### LINKED ISSUE #362: Document how to configure OpenShift with custom image tag, registry etc

Preeticp:
Adding a new section for Configuring Services. 

Fix for issue #362

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -53,9 +53,8 @@ https://github.com/projectatomic/adb-atomic-developer-bundle[ADB]
 vagrant box, from the
 https://github.com/projectatomic/adb-atomic-developer-bundle/tree/master/components/centos[repository].
 For further details on the usage of custom Vagrantfiles designed for
-specific use cases, refer to the
-https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/using.rst[Usage
-Documentation].
+specific use cases, refer to the Download ADB section in the
+https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/installing.adoc[Installing ADB] Documentation.
 
 1.  Start the ADB vagrant box using `vagrant up`. For detailed
 instructions consult the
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -170,6 +166,29 @@ config.servicemanager.http_proxy_password = <Proxy user password>
...
+[NOTE]
+====
+- Docker is the default service for the Atomic Developer Bundle and does not require any configuration to ensure it is started whereas, the Red Hat Enterprise Linux Container Development Kit, which is based on ADB, automatically starts OpenShift. +
+- You can use a comma-separated list to enable multiple services. For instance: docker, openshift.
+====
+
+. Enable the relevant option for the services you have selected in the Vagrantfile. For example, specific versions of OpenShift can be set using the following variables:
++
+
+- `config.servicemanager.openshift_docker_registry = "docker.io"` - Specifies the registry from where the service should be pulled.
```
hferentschik: This is not related to OpenShift afaik. This is really a Docker service configuration option. To use a custom OpenShift version you need 'openshift_image_name' and 'openshift_image_tag'. @budhrg, correct?

@Preeticp Any chance you actually tested this? Is this still working end to end?

README.adoc
```diff
@@ -170,6 +166,29 @@ config.servicemanager.http_proxy_password = <Proxy user password>
...
++
+- `config.servicemanager.openshift_image_name = "openshift/origin"` - Specifies the image to be used.
++
+- `config.servicemanager.openshift_image_tag = "v1.2.0"` - Specifies the version of the image to be used.
```
hferentschik: maybe 1.3.0 instead of 1.2.0 ( a bit more recent)

hferentschik (changes requested):
Looking good, except for the 'openshift_docker_registry' option which should be listed somehow separately as an option for customizing the Docker service.

README.adoc
brgnepal: @Preeticp yes we can change to 1.3.0

README.adoc
praveenkumar: @hferentschik it is related to openshift if openshift images are coming from different hub than `docker.io` like for our CDK it's from `access.redhat.com` and if someone have private registry which contain those images then using that registry.

hferentschik: @praveenkumar, ok, it might be worth to elaborate on this a bit though.
@praveenkumar does this registry setting only apply for the OpenShift image? I would not think so, right? Is this option just a away to add an additional registry? Can you specify more than one?

Another thing,
"from where the service should be pulled." => "from where the _OpenShift image_ is pulled."

praveenkumar: > does this registry setting only apply for the OpenShift image? 

yes that's right.

> I would not think so, right? Is this option just a away to add an additional registry?

No to add another registry to docker we have different script https://github.com/projectatomic/adb-utils/blob/master/utils/add_insecure_registry.

this option is for "where OpenShift image pulled from".

hferentschik (approved):

PR #438, opened 2016-11-11
Fix #389 Support the use of install-cli in HTTP proxy environment

### LINKED ISSUE #389: Support the use of install-cli in HTTP proxy environment

brgnepal:
Fix #389 

cc @hferentschik


test/test_helper.rb
```diff
@@ -110,6 +113,58 @@ def fake_machine(options = {})
...
+    super(config)
+  end
+
+  def service(req, res)
+    req.attributes['url'] = @url
+    super(req, res)
+  end
+end
+
+module WEBrick
```
brgnepal: Need to override log to filter log message for authentication request.

```
$ bundle exe rake test
Run options: --seed 175

# Running:

...............................[2016-11-11 14:03:44] INFO  Basic Proxy Realm: user: authentication succeeded.
..............

Finished in 35.503259s, 1.2675 runs/s, 2.9293 assertions/s.

45 runs, 104 assertions, 0 failures, 0 errors, 0 skips
```

lib/vagrant-service-manager/binary_handlers/binary_handler.rb
```diff
@@ -146,6 +147,19 @@ def validate_url
 
       private
 
+      def set_proxy_environment
+        return if @machine.config.servicemanager.proxy.nil?
+
+        env_proxy = @machine.config.servicemanager.proxy
```
hferentschik, line 155: Some comments on what you do here and why would be nice.

lib/vagrant-service-manager/binary_handlers/binary_handler.rb
```diff
@@ -146,6 +147,19 @@ def validate_url
...
+        if @machine.config.servicemanager.proxy_user
+          user = @machine.config.servicemanager.proxy_user
+          password = @machine.config.servicemanager.proxy_password
+          env_proxy = env_proxy.sub(%r{(https?:\/\/)*}, "\\1#{user}:#{password}@")
+        end
+
+        ENV['https_proxy'] = env_proxy
```
hferentschik, line 164: Again, a comment on why you set this variable and why you set it in this format.

lib/vagrant-service-manager/binary_handlers/binary_handler.rb
```diff
@@ -146,6 +147,19 @@ def validate_url
 
       private
 
+      def set_proxy_environment
+        return if @machine.config.servicemanager.proxy.nil?
+
+        env_proxy = @machine.config.servicemanager.proxy
+        if @machine.config.servicemanager.proxy_user
+          user = @machine.config.servicemanager.proxy_user
+          password = @machine.config.servicemanager.proxy_password
```
hferentschik, line 158: So what if password is empty? Does it work? Is this still the required syntax? Personally, I'd decouple the creation of the proxy string from setting it. You make it harder to test. If they are separate function you can test just the creation of the _env_proxy_ with various values. A true unit test. This test does both things together and make testing hard again.

test/test_helper.rb
```diff
@@ -110,6 +113,58 @@ def fake_machine(options = {})
   machine
 end
 
+class FakeHTTPProxyServer < WEBrick::HTTPProxyServer
```
hferentschik: I suggest we put this into its own class. test_helper has already so much stuff and we might need this later for feature tests as well.

test/test_helper.rb
hferentschik: Is this not a matter of how you configure your logging? It seems it defaults to stdout. Just provide a logger. Either another StringIO or even the same. Monkey patching the log method seems bad.

test/vagrant-service-manager/proxy_test.rb
```diff
@@ -0,0 +1,109 @@
...
+          @proxy_server.shutdown
+        end
+
+        it 'should send client binary download request' do
+          @handler.build_download_url
+          @handler.validate_url
+          @handler.build_archive_path
+          @handler.ensure_binary_and_temp_directories
+
+          @proxy_thread = Thread.new do
```
hferentschik: So why do you do this in the test and not in setup. This blows up the testing code and makes it hard to see what the actual test is about.

test/vagrant-service-manager/proxy_test.rb
```diff
@@ -0,0 +1,109 @@
...
+            @configs[:port] = 7000
+            @proxy_server = fake_proxy_server(@handler.url, @string_stream, @configs)
+            @proxy_server.start
+          end
+          @machine.config.servicemanager.proxy = 'http://localhost:7000'
+
+          begin
+            @download_thread = Thread.new do
```
hferentschik: -1, why would you need another thread here? The main thread can do the download here, right?

test/vagrant-service-manager/proxy_test.rb
```diff
@@ -0,0 +1,109 @@
...
+              @handler.download_archive
+            end
+          rescue StandardError => e
+            raise e.message
+          end
+
+          @string_stream.truncate(0)
+          sleep 3 # Require for @download_thread to start request
```
hferentschik: -1 absolutely not. Thread synchronization is not done via sleep. But given that the download thread is not even needed, I guess this goes away

test/vagrant-service-manager/proxy_test.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      describe 'Unauthenticated proxy environment' do
+        before do
+          @string_stream = StringIO.new
+          @options = { type: :docker, '--cli-version' => '1.10.0' }
+          @handler = ADBDockerBinaryHandler.new(@machine, @machine.env, @options)
+        end
+
+        after do
+          Thread.kill @download_thread
```
hferentschik: -1 see below

test/vagrant-service-manager/proxy_test.rb
```diff
@@ -0,0 +1,109 @@
...
+          Thread.kill @proxy_thread
+          @proxy_server.shutdown
```
hferentschik: Would that even work? You already killed the thread running the proxy. I doubt this will result in a clean shutdown. You most likely don't even need to kill the thread. If you shutdown the server, the thread will exit and then you  just join with the main thread in the tear down method.

test/vagrant-service-manager/proxy_test.rb
```diff
@@ -0,0 +1,109 @@
...
+              @handler.download_archive
+            end
+          rescue StandardError => e
+            raise e.message
+          end
+
+          sleep 3 # Require for @download_thread to start request
+
+          expected_result = "CONNECT get.docker.com:443 200 URL=#{@handler.url}"
+          @string_stream.string.must_include expected_result
```
hferentschik: This is all a bit convoluted. Why do we not try to encapsulate this all a bit better by pushing more functionality into the proxy. So when we create the proxy, we pass it a list of expected URLs. Then the proxy creates its own StringIO instance and we push as much of the config side of things and thread management into this as well. this makes the actual test leaner as well.

Last but not least we have a method on the proxy which is of the form _proxy.assertRequestReceived_ which internally will go through the logs and verify the requests are in there.

Moving forward one can then add things like _proxy.assertRequestReceivedInOrder_ etc. The fake proxy stuff should do as much as possible.

hferentschik (changes requested):

lib/vagrant-service-manager/binary_handlers/binary_handler.rb
brgnepal: > So what if password is empty? Does it work?

No, either it has to be `proxy` or combined `proxy_user` and `proxy_password`. Anyone of `proxy_user`, proxy_password` will fail.

test/vagrant-service-manager/proxy_test.rb
brgnepal: @hferentschik I need the download url to be passed to proxy server. Which I am using as an verifying condition here https://github.com/budhrg/vagrant-service-manager/blob/b2d30e10b28e7a99058a301a588c436fec138bf1/test/test_helper.rb#L123 which is being used in log format https://github.com/budhrg/vagrant-service-manager/blob/b2d30e10b28e7a99058a301a588c436fec138bf1/test/test_helper.rb#L146  .

test/vagrant-service-manager/proxy_test.rb
brgnepal: @hferentschik This step `@handler.download_archive` will download the binary (~30MB) and halt the test.

Hence, I executed this step in separate thread and added sleep of 3 seconds to continue with log verification. 

I hope we are not verifying the download here but the url passed through proxy.

test/vagrant-service-manager/proxy_test.rb
brgnepal: @hferentschik Please check above.

test/test_helper.rb
brgnepal: > Just provide a logger.

Tried that solution but it seems to be not working. Weird behavior is it is only happening with authenticated proxy server.
See the other logs are going to StringIO instance.

test/vagrant-service-manager/proxy_test.rb
brgnepal: > we have a method on the proxy which is of the form proxy.assertRequestReceived

Where @hferentschik ?

lib/vagrant-service-manager/binary_handlers/binary_handler.rb
hferentschik: But we verify this somewhere else, right?

test/vagrant-service-manager/proxy_test.rb
hferentschik: Have the proxy server accept "expected urls" via a setter of some sorts. It keeps a list of expected URLs and when asked for it (via _assertExpectedRequestsReceived_) it checks this list against the log. In fact the log is in this case not even needed. You could check each URL as it arrives and mark expected URLs as received as you go. 

Also, you want to be able to do something like _expectUrlAndReturnOctetStream(url, path_to_file)_. There you say that if you receive a given URL, instead of letting it pass through, create your own HTTP response and return the content of the specified file as data. This way we can avoid network roundtrips for downloading binaries like _oc_.

test/vagrant-service-manager/proxy_test.rb
hferentschik: I see now where you are coming from. You effectively kill the download. I find this flaky. A better solution is the one I mentioned above where the proxy is returning the expected data from a file. Or let it return an error code. This will let the download method naturally fail and you still can assert the proxy request.

test/test_helper.rb
hferentschik: > Tried that solution but it seems to be not working. 

Have you followed the call chain. If the IO instance makes it into the WebRick class, everything looks as it should be. The question is why it does not end up there? If you attach the debugger, you should get the answer pretty much immediately.

test/test_helper.rb
```diff
@@ -110,6 +113,58 @@ def fake_machine(options = {})
...
+    Logger: WEBrick::Log.new(File.open(File::NULL, 'w')),
+    AccessLog: [[stream, '%m %U %s URL=%{url}n']]
+  }
+
+  if options[:authenticated]
+    htpasswd = WEBrick::HTTPAuth::Htpasswd.new options[:htpasswd_file]
+    htpasswd.set_passwd 'Proxy Realm', 'user', 'password'
+    htpasswd.flush
+    # Authenticator
+    authenticator = WEBrick::HTTPAuth::ProxyBasicAuth.new(
```
hferentschik: Here is the cause for your logging issues. An Authenticator also takes a logger (see WEBrick::HTTPAuth::Authenticator) The config of the server is fine and goes to /dev/null resp. the StringIO instance, but since you don't pass a logger to the Authenticator stderr is used there.

Tip: If you run the test with the debugger and break in your custom patch for the log method, you get the full stack trace where this log message originates and it is straight forward from there.

brgnepal:
@hferentschik Updated the PR with the implementation we discussed.

Can't able to get rid of `.......++++` log while running the test

```
$ bundle exec rake test
Run options: --seed 55954

# Running:

.......................++++++
..++++++
.........................++++++
.........................++++++
.....................

Finished in 22.987628s, 1.9576 runs/s, 4.5242 assertions/s.

45 runs, 104 assertions, 0 failures, 0 errors, 0 skips
```

hferentschik:
test this please

hferentschik:
I created follow up pull request #439. I felt we kept talking past each other regarding the design of this.

hferentschik:
Superseded by #439

PR #451, opened 2017-01-23
Fix #446 Support folder as path in install-cli '--path' option

### LINKED ISSUE #446: Support path as folder path instead of path with binary name in install-cli --path option.
Currently, the 'install-cli' --path option need the mention of binary name like 'oc' or 'docker'.

Eg: --path /home/user/cli/oc   

for Openshift.

We should support path as folder instead of binary path.

brgnepal:
Fix #446

### DIFF
--- a/commands.adoc
+++ b/commands.adoc
@@ -242,14 +242,14 @@ As per default, binaries are downloaded to
 To install client binary for Docker:
 
 ----------------------------------------------------------------------------------------------------
-$ vagrant service-manager install-cli docker --cli-version 1.9.1 --path /home/johndoe/bin/docker
+$ vagrant service-manager install-cli docker --cli-version 1.9.1 --path /home/johndoe/bin
 # Binary already available at /home/johndoe/bin/docker
 # run binary as:
 # docker <command>
 export PATH=/home/johndoe/bin:$PATH
 
 # run following command to configure your shell:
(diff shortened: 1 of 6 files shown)


hferentschik (commented):
Changes look good to me. They passed the CI tests against CDK and ADB. Tests also passed on my machine.

Only thing I am wondering whether this is something we should do now, given that we want to make minimal changes for the user. 

I am wondering what others things. @LalatenduMohanty @praveenkumar !?

praveenkumar:
> Only thing I am wondering whether this is something we should do now, given that we want to make minimal changes for the user.

I agree. This change is already tested out and passed CI so let's merge this and then we will not put any effort here.

praveenkumar (approved):

praveenkumar (approved):
LGTM

LalatenduMohanty:
Agree with @hferentschik , this is not a blocker and not planned for CDK 2.4 release. However we can merge this and take this for 2.4 as this has passed review and testing.

brgnepal:
> Agree with @hferentschik , this is not a blocker and not planned for CDK 2.4 release. However we can merge this and take this for 2.4 as this has passed review and testing.

:+1:

hferentschik (approved):
Ok, let's go ahead then ...

PR #453, opened 2017-01-31
Fix #448 install-cli fails

### LINKED ISSUE #448: `vagrant service-manager install-cli openshift´ fails
$ vagrant service-manager install-cli openshift
undefined method `upcase' for nil:NilClass

vagrant-service-manager is at version 1.4.1.

brgnepal:
Fix #448 

I suspect due to change in [vagrant 1.9.1 for winrm communicator](https://github.com/mitchellh/vagrant/commit/725824e1dd3c2225a6274df074142621751bfb14) we are having issues with Windows OS here.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -40,6 +40,7 @@ Metrics/AbcSize:
   Max: 20
   Exclude:
     - 'lib/vagrant-service-manager/command.rb'
+    - 'plugins/guests/redhat/cap/box_version.rb'
     - 'Rakefile'
 
 Lint/Eval:
--- a/lib/vagrant-service-manager/plugin_util.rb
+++ b/lib/vagrant-service-manager/plugin_util.rb
@@ -111,6 +111,7 @@ module VagrantPlugins
 
       def self.execute_once(machine, ui, command)
         machine.communicate.sudo(command) do |_, data|
+          next if data.chomp.empty?
           PluginLogger.debug
           return data.chomp
         end
--- a/plugins/guests/redhat/cap/box_version.rb
+++ b/plugins/guests/redhat/cap/box_version.rb
@@ -15,6 +15,7 @@ module VagrantPlugins
               exit 126
             end
 
+            next if data.chomp.empty?
             return data.chomp if options[:script_readable]
             info = Hash[data.delete('"').split("\n").map { |e| e.split('=') }]
             return "#{info['VARIANT']} #{info['VARIANT_VERSION']}"
--- a/plugins/guests/redhat/cap/machine_ip.rb
+++ b/plugins/guests/redhat/cap/machine_ip.rb
@@ -9,6 +9,7 @@ module VagrantPlugins
 
           PluginLogger.debug
           machine.communicate.execute(command) do |type, data|
+            next if data.chomp.empty?
             ip << data.chomp if type == :stdout
             return "IP=#{ip}" if options[:script_readable]
           end
--- a/plugins/guests/redhat/cap/os_variant.rb
+++ b/plugins/guests/redhat/cap/os_variant.rb
@@ -7,6 +7,7 @@ module VagrantPlugins
           # TODO: execute efficient command to solve this
           return unless machine.communicate.test(command) # Return nil if command fails
           machine.communicate.execute(command) do |_, data|
+            next if data.chomp.empty?
             return data.chomp.delete('"').split('=').last
           end
         end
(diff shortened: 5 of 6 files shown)


.rubocop_todo.yml
```diff
@@ -40,6 +40,7 @@ Metrics/AbcSize:
   Max: 20
   Exclude:
     - 'lib/vagrant-service-manager/command.rb'
+    - 'plugins/guests/redhat/cap/box_version.rb'
```
hferentschik, line 43: why do we have to violate the style and an exception here?

lib/vagrant-service-manager/plugin_util.rb
```diff
@@ -111,6 +111,7 @@ def self.execute_and_exit_on_fail(machine, ui, command)
 
       def self.execute_once(machine, ui, command)
         machine.communicate.sudo(command) do |_, data|
+          next if data.chomp.empty?
```
hferentschik, line 114: So what are you doing here and why. What is the root cause?

hferentschik (changes requested):
So what is the root cause here?

.rubocop_todo.yml
brgnepal: @hferentschik Add additional `if` increase the [Abc size](http://www.rubydoc.info/gems/rubocop/0.27.0/RuboCop/Cop/Metrics/AbcSize) by 0.60 (20.60/20).

Condition is now required to check whether data is empty or not  as from vagrant 1.9.1 this block execute twice for Windows OS only. I suspect due to [change in vagrant 1.9.1 for winrm communicator](https://github.com/mitchellh/vagrant/commit/725824e1dd3c2225a6274df074142621751bfb14).

One solution could be to remove map which might have increase Abc size here.

lib/vagrant-service-manager/plugin_util.rb
brgnepal: See https://github.com/projectatomic/vagrant-service-manager/pull/453#discussion_r99331701

.rubocop_todo.yml
hferentschik: > Condition is now required to check whether data is empty or not as from vagrant 1.9.1 this block execute twice for Windows OS only

The command is executed twice? That sounds like a serious bug. Is there something in Vagrant?

lib/vagrant-service-manager/plugin_util.rb
hferentschik: So why _next_ and not _return_. This seems inconsistent with the rest of the code. The aim is to return a string, right? Would that not by side effect return _true_ or something.

.rubocop_todo.yml
hferentschik: Is Vagrant 1.9.1 our target Vagrant version?

brgnepal: No @hferentschik , we fallback to 1.8.1 and Virtualbox 5.0.26 (CDK 2.3) due to issues in vagrant 1.9.1 in Windows and other issues reported by QE (also updated in mail AFAIR).

lib/vagrant-service-manager/plugin_util.rb
brgnepal: @hferentschik , to give more insights here, when we run command with `machine.communicate.sudo(command)` and use block format as done here, we basically get 
stream type( `_` ignored here) and data executed by that command. More info [here](https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/plugin/v2/communicator.rb#L94-L102).

Now, in case of Windows OS, the `data` part contains expected info on 2nd iteration which was not the case in 1.8.1 (didn't check later versions). In first iteration, `data` part is empty and hence `next`.
Hope it make sense.

hferentschik: > Now, in case of Windows OS, the data part contains expected info on 2nd iteration 

This is the part I don't get, where is the loop here? There is no loop, right?

hferentschik: >  In first iteration, data part is empty and hence next.

Not really. Can you show me the loop?

hferentschik: It seems we are missing the root cause here. No way that this gets executed twice by design. Either there is a bug in Vagrant or we are missing something else.

If there is indeed something we need to workaround we need a better patch than this. What if for some reason we need another communicator call somewhere else? How would one kn ow/remember to add this _next_ statement. If there is a workaround needed we need to provide a way to apply this workaround to all potential uses of the communicator, for example by adding our own helper/utility method around the communicator here. 

But to start with we need to understand the problem first.

brgnepal: Got you

brgnepal: Created https://github.com/mitchellh/vagrant/issues/8259 expecting any reply from vagrant folks. I will debug it in my free time considering other priorities.

brgnepal:
Closing this PR since it actually caused due to vagrant version 1.9.1.

PR #455, opened 2017-02-03
Fix #454 fix error message for --path option

### LINKED ISSUE #454: install-cli command --path option show weird error when the directory is absent
When running `install-cli` with `--path` option to specify folder which doesn't exist then it shows weird error:
```
$ vagrant service-manager install-cli openshift --path /home/budhram/bin/unknown_dir
undefined local variable or method `dir_name' for #<VagrantPlugins::ServiceManager::Installer:0x00000003a49900>
```

brgnepal:
Fix #454 

Unfortunately missed.

### DIFF
--- a/lib/vagrant-service-manager/installer.rb
+++ b/lib/vagrant-service-manager/installer.rb
@@ -45,7 +45,8 @@ module VagrantPlugins
 
         # return if --path is not specified
         return unless @options.key?('--path')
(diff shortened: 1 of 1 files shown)


hferentschik (changes requested):
What's about a test here. Can we not add a test showing this works now as expected? Either unit or integration test?

